PR #3, opened 2023-10-16
Explicitly type final as any[]

CAnthonySiever:
The issue that you were running into with `final` implicitly having a type of `any[]` was likely it not being explicitly `any[]`.

I don't know if this is the sort of error that would prevent it from being able to build locally, but I was able to do so.


CAnthonySiever:
Ah, looks like Vercel handles checking that for us. 😅

PR #7, opened 2023-11-06
Feature/show eliminated

jjm3x3:
This is a kind of "hacky" PR that I just wanted to try out. This further leverages web scraping in order to determine which teams have been eliminated. This code looks pretty bad so I think this is me learning my lesson and next time I want to pull something from Wikipedia, I will make sure that I use the API.

One other nice to note thing though is this will setup and auto refresh to scrap Wikipedia every hour so the site should have up to date info w/in the hour.

### DIFF
--- a/app/contestants/page.tsx
+++ b/app/contestants/page.tsx
@@ -35,7 +35,7 @@ async function getData() {
     return { props: { runners: final } }
 }
 
-function getIsParticipating(item) {
+function getIsParticipating(item: any) {
 
     var row = null
     if (item !== null &&
--- a/app/contestants/page.tsx
+++ b/app/contestants/page.tsx
@@ -49,7 +49,13 @@ export default async function Contestants() {
           <p className="text-lg text-center">{final.props.runners.length} teams</p>
           <br/>
           <div className="text-center">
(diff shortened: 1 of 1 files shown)


app/contestants/page.tsx
```diff
@@ -4,14 +4,14 @@ const wikiUrl = "https://en.wikipedia.org/wiki/The_Amazing_Race_35"
...
-    const response = await fetch(wikiUrl)
+    const response = await fetch(wikiUrl, {next: { revalidate: 3600 }})
     const responseText = await response.text()
     const doc = new JSDOM(responseText)
     var domQuery = doc.window.document.querySelectorAll(".vcard .fn")
 
     // Build data model
     const final = []
-    var team = ""
+    var team = {teamName: "", isParticipating: true}
```
CAnthonySiever, line 14: Would it be better to set the default value here for `isParticipating` to `false`?

jjm3x3: Maybe, my thought was that every team starts the game participating so in a way, that is the default. But I am open to a counter.

CAnthonySiever (approved):

CAnthonySiever:
Definitely a bit hacky, but about as good as it can get with this method 😅

PR #9, opened 2023-11-11
Feature/add scoring

jjm3x3:
This will add scoring for a single hard coded player which is compared to a perfect score for the league. This will be built off of setting up a "perfect scoring" in #12 and may also come after sprucing up our testing strategy so we can be more confident in our changes. #17 

_Note: this doesn't work currently locally or otherwise for the following_
- [x] the hard coded team names don't know about the change to team names defined by both contestants full names (see the new test in #17 which asserts that)

_Note2: given #25 has been filed which suggests that we should re-evaluate the whole paradigm, I am trying not to let that strictly block this effort since this gets a bit closer to the vision for the site_

### DIFF
--- a/__tests__/models/Team.js
+++ b/__tests__/models/Team.js
@@ -82,3 +82,47 @@ describe('Team isInPlay', () => {
         }
     })
 })
+
+describe('Team static getKey', () => {
+    it('should produce the same result reguardless of the order of the team contestant', () => {
+
+        // Arrange
+        const orderingOne = "Aname one & Bname two"
+        const orderingTwo = "Bname two & Aname one"
+
+        // Act
+        const resultOne = Team.getKey(orderingOne)
+        const resultTwo = Team.getKey(orderingTwo)
+
+        // Assert
+        expect(resultOne).toBe(resultTwo)
+    })
+
+    it('should produce the same result ignoring trailing whitespace', () => {
+
+        // Arrange
+        const expectedTeamName = "Aname one & Bname two"
+        const testTeamName = "Aname one & Bname two   "
+
+        // Act
+        const expectedResult = Team.getKey(expectedTeamName)
+        const testResult = Team.getKey(testTeamName)
+
+        // Assert
+        expect(testResult).toBe(expectedResult)
+    })
+
+    it('should produce the same result ignoring leading whitespace', () => {
+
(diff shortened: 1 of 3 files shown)


app/scoring/page.tsx
```diff
@@ -5,13 +5,37 @@ import TeamList from '../components/teamList'
 import Team from '../models/Team'
 import { shouldBeScored } from '../utils/teamListUtils'
 
+interface Dictionary<T> {
+    [Key: string]: T;
+}
+
+function getKey(teamName: string): string {
```
internette, line 12: Move this function to the team object

jjm3x3: So I really like this suggestion, heck we work-shopped it together in-person, but after trying it out I realized that while this looks good on paper there are two uses of this function and the second one there is only a string and not a full team in order to call this method on. This leads to code that looks like this:
```
        const foundTeam = teamDictionary[(new Team({teamName: x})).getKey()]
```
Which doesn't quite feel right 🤔 

Also, trying to get too deep into this question seems like we are trying to solve #25 which I was really trying to avoid in this PR! 😅

jjm3x3: _update: so I learned that Javascript does have static functions so I tried out that and that wasn't half bad so I added those commits, would love to know what you think!_

app/scoring/page.tsx
```diff
@@ -5,13 +5,25 @@ import TeamList from '../components/teamList'
 import Team from '../models/Team'
 import { shouldBeScored } from '../utils/teamListUtils'
 
+interface Dictionary<T> {
+    [Key: string]: T;
+}
```
CAnthonySiever, line 10: Now this is a bit interesting, but I assume necessary 😅

CAnthonySiever (approved):

__tests__/models/Team.js
```diff
@@ -82,3 +82,47 @@ describe('Team isInPlay', () => {
         }
     })
 })
+
+describe('Team static getKey', () => {
+    it('should produce the same result reguardless of the order of the team contestant', () => {
```
jjm3x3, line 87:
```suggestion
    it('should produce the same result regardless of the order of the team contestant', () => {
```

PR #10, opened 2023-11-12
Feature/create fetch util

jjm3x3:
This looks like a very simple PR, because it is. I am just submitting this so when I make changes to the util after this it is easier to view the diff.


CAnthonySiever (approved):
Makes sense moving that off of `page.tsx` 😅

internette (approved):

app/utils/wikiScraping.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+function getIsParticipating(item: any) {
+
+    var row = null
+    if (item !== null &&
+        item.parentElement !== null &&
+        item.parentElement.parentElement !== null &&
+        item.parentElement.parentElement.parentElement !== null &&
+        item.parentElement.parentElement.parentElement.parentElement !== null) {
```
internette, line 46: I don't like this, but approving anyways because I can't think of something better. If I think of something, I'll open an issue or PR

PR #11, opened 2023-11-12
Feature/accquire eliminationorder

jjm3x3:
Most of the changes here are pretty self explanitory, but just to make sure they get listed for an at a glance look
- Add an `eliminationOrder` prop
- Default the `eliminationOrder` to 0 (would like to make it teams.length, but we don't have that value upon initialization, may be worth trying to clean up after)
- Create a few helpers to that I can get a bit more typing
- One function to get the fullTeamStatus
- Two more helpers to get `isParticipating` and another to get `eliminationOrder`

### DIFF
--- a/app/utils/wikiScraping.tsx
+++ b/app/utils/wikiScraping.tsx
@@ -53,7 +53,7 @@ function getFullTeamStatus(item: any): string {
     return row.lastElementChild.textContent
 }
 
-function getIsParticipating(teamStatusFull: string): bool {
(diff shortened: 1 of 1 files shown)


CAnthonySiever (approved):
I'm not a fan of getting the full team status then plugging that in to a function to determine if people are participating and another function to get their elimination order, but it's about the best you can hope for with processing the data from that endpoint 😅

internette (approved):

PR #12, opened 2023-11-14
Feature/add perfect scoring

jjm3x3:
Here is a change that will show a perfect scoring. This will set the stage for showing an individuals fantasy league contestant's scoring.

_Edit: I did try to see if it would make sense to set eliminationOrder to the max, but this revealed a defect based on a assumption that I was making before and that lead me to want to add some testing to help protect against strange regressions like this see #17 and in particular [this commit/change](https://github.com/jjm3x3/AmazingRaceFantasy/pull/17/commits/ee0e0b9e7c4aecf90993b4295bc4ca5100e1a17c)_

### DIFF
--- a/app/scoring/page.tsx
+++ b/app/scoring/page.tsx
@@ -36,8 +36,8 @@ export default async function Scoring() {
             <h1 className="text-2xl text-center">Current Scoring for {currentSelectedContestant}</h1>
             <br/>
             <div className="text-center">
-                {roundScores.map((s,roundNumber) => {
-                    grandTotal += s
+                {roundScores.map((score, roundNumber) => {
+                    grandTotal += score
                     currentWeek++
 
                     return (<Fragment key={"round details"+roundNumber}>
@@ -50,7 +50,7 @@ export default async function Scoring() {
                             </Fragment>)
                         })}
                         <br/>
-                        <p key={"weekTotal"+roundNumber}className="text-center">Weekly Total: {s}</p>
+                        <p key={"weekTotal"+roundNumber}className="text-center">Weekly Total: {score}</p>
                         <p key={"grandTotal"+roundNumber}className="text-center">Grand Total: {grandTotal}</p>
                         <br/>
                         <br/>
--- a/app/scoring/page.tsx
+++ b/app/scoring/page.tsx
@@ -43,11 +43,11 @@ export default async function Scoring() {
                     return (<Fragment key={"round details"+roundNumber}>
                         <h2 key={"weekHeader"+roundNumber}className="text-xl">Week {currentWeek}</h2>
                         {reverseTeamsList.map(t => {
-                            return (<>
-                                <p key={t.teamName}>
+                            return (<Fragment key={"teamStanding"+t.teamName+roundNumber}>
+                                <p key={t.teamName+roundNumber}>
                                     {t.eliminationOrder === 0 || currentWeek < t.eliminationOrder ? t.teamName : <s>{t.teamName}</s>}
                                 </p>
-                            </>)
+                            </Fragment>)
                         })}
                         <br/>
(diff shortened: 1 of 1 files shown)


app/scoring/page.tsx
```diff
@@ -1,9 +1,59 @@
-export default function Scoring() {
-    var currentSelectedContestant = "Jacob"
+import { getData, wikiUrl } from "../utils/wikiScraping"
 
+export default async function Scoring() {
+
+    const wikiData = await getData()
```
CAnthonySiever, line 5: I believe this conflicts with the API rework changes that have been pulled.

jjm3x3: LOL, it absolutely does, plus at this point there is some unit testing stuff I would like to pull in, in addition to a slight rework of the aforementioned API rework 😓

jjm3x3: Alright, this should be addressed now in #21  and the deal before it (rework of the aforementioned Api rework), #20 and so this code is no longer up to date and has been replaced by code that works.

app/scoring/page.tsx
```diff
@@ -1,9 +1,61 @@
...
+            <div className="text-center">
+                {roundScores.map(s => {
+                    grandTotal += s
+                    currentWeek++
+
+                    return (<>
+                        <h2 className="text-xl">Week {currentWeek}</h2>
+                        {reverseTeamsList.map(t => {
+                            return (<>
+                                <p key={t.teamName}>
```
CAnthonySiever, line 46: Are there going to be duplicate keys here, or are we planning to show scoring for one contestant at a time?

app/scoring/page.tsx
```diff
@@ -1,9 +1,61 @@
...
+
+    const reverseTeamsList = [...pageData.props.runners].reverse()
+    let grandTotal = 0
+    let currentWeek = 0
     return (
         <div>
             <h1 className="text-2xl text-center">Current Scoring for {currentSelectedContestant}</h1>
+            <br/>
+            <div className="text-center">
+                {roundScores.map(s => {
```
CAnthonySiever, line 38: Personal preference, but I kinda like writing out `score` here, especially since line 47 uses `<s>` to strike-through text. Using `s` is perfectly valid here, though, just wanted to float the idea.

CAnthonySiever (approved):

app/scoring/page.tsx
jjm3x3: Totally love that, will make the change. I generally start with just using short variables for lambdas, but this is quite a big lambda so 👍

app/scoring/page.tsx
```diff
@@ -1,9 +1,61 @@
...
+                    grandTotal += s
+                    currentWeek++
+
+                    return (<>
```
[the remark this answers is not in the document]
CAnthonySiever: Oh my god. For some reason I thought this was a react project. 🙃
Sorry, was just looking at some react stuff.

app/scoring/page.tsx
jjm3x3: Well, if by contestant, you mean league contestant, then "yes" we will only show one at a time. But I was looking more into this and this happens on every round rendered to screen from the `roundScores.map()` so based on that, yeah this should be updated to make sure keys are unique across the whole page. Fixing now. 👍

app/scoring/page.tsx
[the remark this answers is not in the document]
jjm3x3: Oh no, you are all good here. Next is built on top of React AFAIU, so that is perfectly reasonable mistake to make, plus I think this suggestion is not only a good one but actually a necessary one since both this fragment and the one below should have a `key` prop and I think the only way to do that and not confuse the heck out of the compiler is to clearly mark them as `Fragment`s

CAnthonySiever (approved):

CAnthonySiever (approved):

Issue #5, opened 2023-11-01
Use Wikipedia  Api


jjm3x3:
In order to avoid site breakages when the Wikipedia site changes we should switch to use the Wikipedia Api. Once this is completed all references to Jsdom should be removed (both the core lib and the types lib)

jjm3x3:
So @CAnthonySiever this work was done to the best of your abilities it seems, but the stated goal in the issue was not really achieved. Even though it is a slight change up in the approach I am wondering if it makes sense to take any further conversation on this to #25 WDYT?

CAnthonySiever:
I owe you something that gets the required section number from the Wikipedia API (I know this is possible but haven't had the time to commit to that, hopefully soon 🙏 ), that way we don't have to manually change the section number whenever it moves around, but I believe that will satisfy that. As it is currently, I agree it is only _slightly_ more reliable (if that) when looking to avoid site-breakages whenever the wiki is updated.

PR #13, opened 2023-11-14
Rework the process of pulling data from Wikipedia to utilize its API …

CAnthonySiever:
…and cheerio to address Issue #5 

Let me know if anything needs changing even if its simply to fit an existing paradigm.

There's still a bit of the fun that is pulling from a table in Wikipedia involved, but  I tried to keep it as clean as possible.

Also, since we're pulling data from the table, I've included (read: kept) the `age`, `relationship`, and `hometown` fields, but haven't changed any of the page to make use of this data. I can remove this if we feel it's unnecessary.


app/utils/wikiQuery.tsx
```diff
@@ -0,0 +1,75 @@
...
+    $('table.wikitable tbody tr').each((index, element) => {
+
+        const $row = $(element)
+        const teamName = $row.find('th span.fn').text().trim()
+        const age = parseInt($row.find('td').eq(0).text().trim(), 10)
+        const relationship = $row.find('td').eq(1).text().trim()
+        const hometown = $row.find('td').eq(2).text().trim()
+        const status = $row.find('td').eq(3).text().trim()
+
+        if ( index %2 == 1) {
```
internette, line 44: Nit: Whitespacing

internette:
@CAnthonySiever  I'll wait for @jjm3x3 to weigh in, but I know he and I bounced the idea around him creating a JSON file instead.

PR #16, opened 2023-11-22
 Rework the process of pulling data from Wikipedia to utilize its API

CAnthonySiever:
…and cheerio to address Issue https://github.com/jjm3x3/AmazingRaceFantasy/issues/5

Let me know if anything needs changing even if its simply to fit an existing paradigm.

There's still a bit of the fun that is pulling from a table in Wikipedia involved, but I tried to keep it as clean as possible.

Also, since we're pulling data from the table, I've included (read: kept) the age, relationship, and hometown fields, but haven't changed any of the page to make use of this data. I can remove this if we feel it's unnecessary.

This is a duplicate PR of #13  after addressing some issues of case-sensitivity and insensitivity within github 😅


CAnthonySiever:
_Originally posted by @internette in https://github.com/jjm3x3/AmazingRaceFantasy/issues/13#issuecomment-1811027441_1)_
> @CAnthonySiever I'll wait for @jjm3x3 to weigh in, but I know he and I bounced the idea around him creating a JSON file instead.

jjm3x3 (approved):
Not sure I see a strict benefit in switching from JSDOM to Cheerio which AFICT we are using them for roughly the same thing, which is parsing the Dom which comes back from scraping the Wikipedia web page. This change doesn't seem to use the Api as the title and description suggest, but I do like the use of interfaces here and there are some other small changes I am a fan of. Lets move forward with this and we'll take a look at some of the rougher edges going forward.

CAnthonySiever:
> Not sure I see a strict benefit in switching from JSDOM to Cheerio which AFICT we are using them for roughly the same thing, which is parsing the Dom which comes back from scraping the Wikipedia web page. This change doesn't seem to use the Api as the title and description suggest, but I do like the use of interfaces here and there are some other small changes I am a fan of. Lets move forward with this and we'll take a look at some of the rougher edges going forward.

Apologies for the confusion, I left `wikiUrl` in place since it was being exported, and added an `apiUrl` to use in `fetchWikipediaData`, but I can move `wikiUrl` to `pages.tsx` to limit confusion. And I believe we are using JSDOM and Cheerio to do the same thing, but by my understanding Cheerio is a bit more efficient about it. I am open to using either though.

jjm3x3:
> > Not sure I see a strict benefit in switching from JSDOM to Cheerio which AFICT we are using them for roughly the same thing, which is parsing the Dom which comes back from scraping the Wikipedia web page. This change doesn't seem to use the Api as the title and description suggest, but I do like the use of interfaces here and there are some other small changes I am a fan of. Lets move forward with this and we'll take a look at some of the rougher edges going forward.
> 
> Apologies for the confusion, I left `wikiUrl` in place since it was being exported, and added an `apiUrl` to use in `fetchWikipediaData`, but I can move `wikiUrl` to `pages.tsx` to limit confusion. And I believe we are using JSDOM and Cheerio to do the same thing, but by my understanding Cheerio is a bit more efficient about it. I am open to using either though.

Ah, got it! thanks for pointing out that we are using that Wikipedia Url for something else. I would say we don't need to move it just yet, but perhaps consider where a good shared location would be for it.

PR #17, opened 2023-11-26
Feature/add testing

jjm3x3:
_see note..._
Here is a PR that will add jest to unit testing to allow for better testing to make sure regressions don't come up. In order to run this in the Vercel platform the deployment command must be updated. At this time I have opted to update it from `next build` to `jest && next build` which seems to work good enough.

_note: the tests were already added in #21 but here is the addition of some more tests for code that is already written_

Now that the focus of this PR has changed every so slightly a few things to look out for.
- [x] This currently assumes that the first player/contestant in the list of contestants will be empty (@CAnthonySiever and I have already discussed adding a small filtering function to address this perhaps done in this PR)
  - _update: I started taking a look at this and it doesn't seem super difficult, but outside the scope of this which was just to add some tests on existing things so I have gone ahead and opened issue #24_
- [x] This will add a test to assert things on team names since getting a fantasy leagues contestants score is currently dependent on it (see #9) It is unclear if this is how team names should be constructed and how these two inputs (one from wiki and one from fantasy league contestant) should be matched
  - _update: Upon thinking and looking more into this I am thinking that we should proceed assuming that we will use both contestants full names with the '&' character for now and we can move further discussion into #25_ 
- [x] This PR started on the premise that if a teams eliminationOrder is initialized to the max then it will break the scoring since the scoring will be calculated on the max of all eliminationOrders which is a bit misleading week over week. The test to assert that behavior is here, but the question is that the right assertion or should it be changed.
  - _update: the short answer is "yes." Upon some outside discussion it dawned on me "what do we get my making a teams elimination order start out set to the max" and I couldn't come up with anything, so until we do I think it is fine to assume this behavior and move on, only to reevaluate if we find a good reason to._

### DIFF
--- a/app/utils/wikiFetch.tsx
+++ b/app/utils/wikiFetch.tsx
@@ -35,12 +35,18 @@ export async function getWikipediaContestantData(): Promise<IWikipediaContestant
 
         const $row =  $(element)
 
+        const name = $row.find('th span.fn').text().trim()
+        const age = $row.find('td').eq(0).text().trim()
+        const relationship = $row.find('td').eq(1).text().trim()
+        const hometown = $row.find('td').eq(2).text().trim()
+        const status = $row.find('td').eq(3).text().trim()
+
         const aContestant: IWikipediaContestantData = {
-            name: $row.find('th span.fn').text().trim(),
-            age: $row.find('td').eq(0).text().trim(),
-            relationship: $row.find('td').eq(1).text().trim(),
-            hometown: $row.find('td').eq(2).text().trim(),
-            status: $row.find('td').eq(3).text().trim()
+            name: name,
+            age: age,
+            relationship: relationship,
+            hometown: hometown,
+            status: status
         }
 
         return aContestant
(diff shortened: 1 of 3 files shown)


CAnthonySiever (approved):
This looks like it was a headache to sort out 😅 But I'm very glad you did. 
It'll be nice to have some tests to lean on as more features are implemented.

app/utils/wikiFetch.tsx
```diff
@@ -35,12 +35,18 @@ export async function getWikipediaContestantData(): Promise<IWikipediaContestant
 
         const $row =  $(element)
 
+        const name = $row.find('th span.fn').text().trim()
+        const age = $row.find('td').eq(0).text().trim()
+        const relationship = $row.find('td').eq(1).text().trim()
+        const hometown = $row.find('td').eq(2).text().trim()
+        const status = $row.find('td').eq(3).text().trim()
```
internette, line 42: I love this refactor. Beautiful and legible. Nice job

internette (approved):

CAnthonySiever (approved):

PR #18, opened 2023-11-26
Bugfix/update package lock

jjm3x3:
This got missed in the most recent PR, lets make sure this gets included in all patches, it gives us reproduceable builds so we can keep folks from needing to track down strange dependency issues.


CAnthonySiever (approved):
Nothing wild is sticking out, and I'm certain any mistakes would have showed up when running `npm i`👍

PR #19, opened 2023-12-10
Bugfix/Update Wiki Section

jjm3x3:
Update section to make sure we are pulling contestants

So two takeaways
1. We probably can't rely on wikipedia "sections" 
1. So long as we continue to get data this way, we should consider how we can get notified of the site breaking 😬

### DIFF
--- a/app/utils/wikiQuery.tsx
+++ b/app/utils/wikiQuery.tsx
@@ -19,7 +19,7 @@ export interface IContestant {
 }
 
 async function fetchWikipediaData(): Promise<IWikipediaData> {
-    const apiUrl = 'https://en.wikipedia.org/w/api.php?action=parse&format=json&page=The_Amazing_Race_35&section=5&formatversion=2'
+    const apiUrl = 'https://en.wikipedia.org/w/api.php?action=parse&format=json&page=The_Amazing_Race_35&section=7&formatversion=2'
     const response = await fetch(apiUrl)
     const data = await response.json()
     return data


app/utils/wikiQuery.tsx
```diff
@@ -19,7 +19,7 @@ export interface IContestant {
 }
 
 async function fetchWikipediaData(): Promise<IWikipediaData> {
-    const apiUrl = 'https://en.wikipedia.org/w/api.php?action=parse&format=json&page=The_Amazing_Race_35&section=5&formatversion=2'
+    const apiUrl = 'https://en.wikipedia.org/w/api.php?action=parse&format=json&page=The_Amazing_Race_35&section=7&formatversion=2'
```
CAnthonySiever, line 22: This looks correct.
Also as a note: Going to the article and clicking on the `Edit` button above the table will take you to a page where the url will have the section that we want listed as a parameter. In this case this url would be `https://en.wikipedia.org/w/index.php?title=The_Amazing_Race_35&action=edit&section=7`.
image

CAnthonySiever (approved):

app/utils/wikiQuery.tsx
jjm3x3: Ooo, that is helpful. I just used the guess an check method, but this is good to know! I am also wondering how we can be notified of an issue. Since this will refresh it's cache once a day I wonder if it could fire an alert if it pulls some data that looks way out of whack? 🤔

CAnthonySiever: Apologies on the delay, wanted to look at a few things. 
We could probably add an error for when the `line` field in the first item in the `sections` array in the JSON response for our query is not `Cast`.
However, looking at that shows us we could probably query `https://en.wikipedia.org/w/api.php?action=parse&format=json&page=The_Amazing_Race_35&formatversion=2` first to get a list of sections, then query the section based on the index of the section called `Cast`. 
This does not protect us from them changing the title of the section to something else in the future.

CAnthonySiever: Although I assume an `anchor` is probably less subject to change, and the anchor on that section is also conveniently called `Cast`.

jjm3x3: This is a great idea, I totally think this is worth doing. Would even welcome a PR if you wanted to go ahead and update this to work that way, or if not at least a ticket would be nice! Just a heads up I am going to give priority to [this branch](https://github.com/jjm3x3/AmazingRaceFantasy/compare/tmp) which breaks apart the scraping and the parsing a lil bit so that the parsing and aggregating into teams can be tested and so that the parts can change independently.

PR #20, opened 2023-12-11
Feature/seperate scraping and parsing

jjm3x3:
Split up the fetching and some trivial parsing into it's own module so we can write more testable parsing logic when it comes to transforming contestants into teams.


app/utils/wikiFetch.tsx
```diff
@@ -0,0 +1,50 @@
...
+        const aContestant: IWikipediaContestantData = {
+            name: $row.find('th span.fn').text().trim(),
+            age: $row.find('td').eq(0).text().trim(),
+            relationship: $row.find('td').eq(1).text().trim(),
+            hometown: $row.find('td').eq(2).text().trim(),
+            status: $row.find('td').eq(3).text().trim()
+        }
+
+        return aContestant
+    }).toArray()
```
internette, line 47: Nit: I prefer to pull this code out into a function called something like "formatContestantData". But that's more my taste than anything else.

internette (approved):

app/utils/wikiQuery.tsx
```diff
@@ -1,45 +1,20 @@
...
-export const wikiUrl = "https://en.wikipedia.org/wiki/The_Amazing_Race_35"
-
-interface IWikipediaData {
-    parse: {
-        text: string
-    }
-}
-
-export interface IContestant {
+export interface ITeam {
```
CAnthonySiever, line 3: In the future, we should probably move these interfaces, including the two in `wikiFetch.tsx`, to a `models` or `interfaces` file, that way if anything changes they're a bit easier to track down. 

We are absolutely fine to leave this for now, but we may want to mark it as an issue if it's something I can't get to this week.

CAnthonySiever (approved):

app/utils/wikiQuery.tsx
```diff
@@ -1,45 +1,20 @@
...
 
-async function fetchWikipediaData(): Promise<IWikipediaData> {
-    const apiUrl = 'https://en.wikipedia.org/w/api.php?action=parse&format=json&page=The_Amazing_Race_35&section=7&formatversion=2'
-    const response = await fetch(apiUrl)
-    const data = await response.json()
-    return data
-}
-
-export async function getContestantList(): Promise<any> {
+export function getTeamList(contestantData :IWikipediaContestantData[]): any {
```
CAnthonySiever, line 10: We could probably change `any` here to `ITeam[]`.

jjm3x3: The issue with that is it still returns a `return { props: { runners: contestants }}` which is to say it returns some object with a props object which has a runners property which is a `ITeam[]` 😅 and TBH I was just a tad too lazy to create interfaces for all those. But really there is nothing stopping us, you are right!

app/utils/wikiQuery.tsx
jjm3x3: Alright, I am opting to not do this right at the moment, because I just wanna keep development moving and I don't feel like this is a blocker, but I do agree with this premise so I created #22

app/utils/wikiQuery.tsx
jjm3x3: Similar to the above, I going to opt to not do this at the moment, but I do think we should keep this in mind because relying on the any type here is not good!

PR #21, opened 2023-12-11
Feature/setup testing

jjm3x3:
Here are the things necessary to start getting unit testing going.

### DIFF
--- a/__tests__/wikiQuery.js
+++ b/__tests__/wikiQuery.js
@@ -33,7 +33,8 @@ describe('getData', () => {
                 name: firstContestantsFirstName + " Guy",
                 status: "Participating"
             },
-            {team: secondContestantsFirstName + " Brother"}]
+            {team: secondContestantsFirstName + " Brother"}
+        ]
 
         // Act
         var act = () => getTeamList(listOfContestants)
--- a/__tests__/wikiQuery.js
+++ b/__tests__/wikiQuery.js
@@ -22,4 +22,23 @@ describe('getData', () => {
         // Assert
         expect(result).not.toBeNull()
     })
+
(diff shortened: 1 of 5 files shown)


__tests__/wikiQuery.js
```diff
@@ -0,0 +1,44 @@
...
+        // Arrange
+        const firstContestantsFirstName = "Some"
+        const secondContestantsFirstName = "SomeGuys"
+
+        const listOfContestants = [
+            {
+                name: firstContestantsFirstName + " Guy",
+                status: "Participating"
+            },
+            {team: secondContestantsFirstName + " Brother"}]
```
internette, line 36: Formatting

internette (changes requested):

internette (approved):
LGTM

__tests__/wikiQuery.js
```diff
@@ -0,0 +1,45 @@
...
+            {
+                name: secondContestantsFirstName + " Brother",
+                status: "Participating"
+            }]
+
+        // Act
+        var result = getTeamList(listOfContestants)
+
+        // Assert
+        expect(result).not.toBeNull()
```
CAnthonySiever, line 23: I would go so far to say that we're not just expecting a non null result here, but also there to be some specific values regarding some guy and his brother. 

Although, I believe it not being null is certainly sufficient for now.

CAnthonySiever (approved):

__tests__/wikiQuery.js
jjm3x3: I totally agree, I do have other tests coming soon, but I also wanted to get this in ASAP so that in case others start making changes they can use this as a guide for how to add tests to make sure their changes are working!

PR #23, opened 2023-12-12
Bugfix/elimination is number

jjm3x3:
In light of the recent changes of polling the API elimination order was setup as a string and it shouldn't be since we want to order teams based on it.

### DIFF
--- a/__tests__/wikiQuery.js
+++ b/__tests__/wikiQuery.js
@@ -42,4 +42,26 @@ describe('getData', () => {
         // Assert
         expect(act).toThrow(new ReferenceError("Status is either null or undefined and it shouldn not be"))
     })
+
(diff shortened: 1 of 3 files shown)


CAnthonySiever (approved):

__tests__/wikiQuery.js
```diff
@@ -42,4 +42,26 @@ describe('getData', () => {
...
+            },
+            {
+                name: secondContestantsFirstName + " Brother",
+                status: "Participating"
+            }
+        ]
+
+        var result = getTeamList(listOfContestants)
+
+        expect(result.props.runners[0].eliminationOrder).toEqual(2)
```
CAnthonySiever, line 65: 🙌

CAnthonySiever (approved):

Issue #24, opened 2023-12-29
Fix incoming data to filter for empty contestants


jjm3x3:
After reworking the wiki scraping the first time to instead us the Api and use the Cherio library, we started getting an empty contestant in our list which lead to some strange code changes that were necessary. Maintaining this strange behavior seems like an unnecessary burden so the suggestion has been made to add a filter to remove contestants which appear to be empty. Once completed the current [wikiQuery](https://github.com/jjm3x3/AmazingRaceFantasy/blob/5c228c966a51e5d9ea710d3e9e7d440531cba963/app/utils/wikiQuery.tsx) util should not need to account for any of the contestants in the incoming array being empty let alone the first one.

jjm3x3:
It would be my two cents that this is done by making sure that we never get an array which contains empty contestants from what is currently called the [wikiFetch](https://github.com/jjm3x3/AmazingRaceFantasy/blob/5c228c966a51e5d9ea710d3e9e7d440531cba963/app/utils/wikiFetch.tsx#L26) util. It seems like our data utilities should behave better than that so that we can write pages where we can rely on our data.

jjm3x3:
Should be resolved by #26

PR #26, opened 2023-12-30
Feature/filter empty contestants

jjm3x3:
Here is a fairly simple PR to do exactly as it's name implies. This should cover the issue that is described in #24 and mentioned in #17

### DIFF
--- a/app/utils/wikiFetch.tsx
+++ b/app/utils/wikiFetch.tsx
@@ -54,3 +54,8 @@ export async function getWikipediaContestantData(): Promise<IWikipediaContestant
 
     return contestantData
 }
+
+async function filterEmptyContestants(contestantList: IWikipediaContestantData[]): IWikipediaContestantData[] {
+    return contestantList.filter(x => x.name !== "")
+}
+
--- a/app/utils/wikiFetch.tsx
+++ b/app/utils/wikiFetch.tsx
@@ -58,6 +58,6 @@ export async function getWikipediaContestantData(): Promise<IWikipediaContestant
 export function filterEmptyContestants(contestantList: IWikipediaContestantData[]): IWikipediaContestantData[] {
     // I believe that the behavior that led to this addition is missing all
     // properties but the name seemed like the most important one for now
-    return contestantList.filter(x => x.name != null)
+    return contestantList.filter(x =>  x.name != null && x.name !== "")
 }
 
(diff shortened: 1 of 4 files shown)


app/utils/wikiFetch.tsx
```diff
@@ -52,5 +52,14 @@ export async function getWikipediaContestantData(): Promise<IWikipediaContestant
         return aContestant
     }).toArray()
 
-    return contestantData
+    const result = filterEmptyContestants(contestantData)
+
+    return result
+}
+
+export function filterEmptyContestants(contestantList: IWikipediaContestantData[]): IWikipediaContestantData[] {
```
jjm3x3, line 60: This is only exported so that it can be tested. Certainly open to other suggestions here.

app/utils/wikiFetch.tsx
```diff
@@ -52,5 +52,14 @@ export async function getWikipediaContestantData(): Promise<IWikipediaContestant
...
+    // I believe that the behavior that led to this addition is missing all
+    // properties but the name seemed like the most important one for now
+    return contestantList.filter(x =>  x.name != null && x.name !== "")
```
jjm3x3, line 63: I can prove this checks works via the tests, but I have no idea if this is the canonical way to do this kind of thing 😅

app/utils/wikiFetch.tsx
CAnthonySiever: Outside of running tests in this file, I believe you need to export it.

app/utils/wikiFetch.tsx
CAnthonySiever: I believe  `contestantList.filter(x => x.name);`  would also filter out names that are `null` or empty strings.

app/utils/wikiQuery.tsx
```diff
@@ -16,14 +16,15 @@ export function getTeamList(contestantData :IWikipediaContestantData[]): any {
         const status = element.status
         let teamName = element.name
 
-        if (index % 2 == 1) {
+        if (status === null || status === undefined) {
+            throw new ReferenceError("Status is either null or undefined and it shouldn not be")
```
CAnthonySiever, line 20: `shouldn not be`

app/utils/wikiQuery.tsx
```diff
@@ -16,14 +16,15 @@ export function getTeamList(contestantData :IWikipediaContestantData[]): any {
...
+        }
+
+
```
CAnthonySiever, line 23: double-space

CAnthonySiever (approved):

app/utils/wikiFetch.tsx
jjm3x3: Oh? How come? it is only used, right in this module.

CAnthonySiever (approved):

PR #27, opened 2024-01-01
Feature/perfect scoring cleanup

jjm3x3:
This is a pretty simple PR, but this will give a hint about the direction I am hoping we go with the site. Since Next is built on React it only seems fitting to do some modern development and start building sections into components. There are just a few things that include:
- Creating a `TeamList` component. It is meant to be re used when ever a team needs to be listed and possibly have names crossed off
- Remove the `weeklyScore` list since it was entirely unused
- Remove the `currentWeek` variable it was used, but it was only different than the `roundNumber` because `roundNumber` is 0 based and currentWeek was meant to display so it is 1 based. I figured no need for the extra variable.

### DIFF
--- a/app/components/teamList.tsx
+++ b/app/components/teamList.tsx
@@ -1,0 +1,14 @@
+import { Fragment } from 'react'
+
+export default function TeamList({ teamList, roundNumber }) {
+    
+    return <div>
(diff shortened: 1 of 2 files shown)


internette (approved):
Beautiful refactor

Issue #25, opened 2023-12-29
Re-evaluate wiki & leage contestant team tiying


jjm3x3:
Similar to the setup to #24 once we did our first wiki refactor to the Api it broke one assumption that was made that we can just use the "team name" which at the time was just each contestants first first name (you did not misread, we intentionally included two "firsts" there see the read on) joined with spaces and a '&' character. This turned out to have one immediate issue when it comes to contestants with multiple first or multiple last names (AKA a person with 3+ names, how do you know which names are part of their first name). Then after the aforementioned Api refactor another issue came up that if we the team names suddenly change then it will break the logic that ties them together with a league contestants order entirely. In this particular case it went from the previous definition of team name to now included both contestants full names joined with spaces and a '&' character. Potentially this is better since this doesn't suffer as much from the first first name issue described and can be used in the short term, but I still see a few potential issues. In no particular order 
- A persons name could change
- Unclear what name should come first in the pair
- Managing lists of full names is a bit cumbersome and error prone
- We may not want to display every contestant's full name (UI/UX) consideration

Not clear how to address this, or if it _needs_ to be addressed, but here is the place where we can start that discussion...

PR #28, opened 2024-01-06
Feature/add team model

jjm3x3:
This change helps captured what I realized as I was playing around w/ the scoring, (which is getting close btw). What I learned is that there is some rather complicated logic involved in determining if a team should be crossed off and this will also play a role on if it is scored since those go hand in hand. In the interest of trying for a "heavy  model/thin controller" idea here. Since the team is best equipped to know things about itself if we write small methods on it passing it simple arguments those become very well tested pieces of code that can be reused with confidence.

### DIFF
--- a/__tests__/models/Team.js
+++ b/__tests__/models/Team.js
@@ -9,3 +9,22 @@ describe('Team construct', () => {
         expect(aTeam).not.toBeNull()
     })
 })
+
+describe('Team isInPlay', () => {
+    it('should be true when elimination order is 0 reguardless of roundNumber (sparse)', () => {
+        const aTeam = new Team({
+            eliminationOrder: 0
+        })
+
+        // Assert, just check sections because otherwise it takes too long
+        for (let i = Number.MAX_VALUE - 100; i < Number.MAX_VALUE; i++) {
+            expect(aTeam.isInPlay(i)).toBeTruthy()
+        }
+        for (let i = Number.MIN_VALUE + 100; i > Number.MIN_VALUE; i--) {
+            expect(aTeam.isInPlay(i)).toBeTruthy()
+        }
+        for (let i = 0 - 100; i < 100; i++) {
+            expect(aTeam.isInPlay(i)).toBeTruthy()
+        }
+    })
+})
--- a/__tests__/models/Team.js
+++ b/__tests__/models/Team.js
@@ -27,4 +27,20 @@ describe('Team isInPlay', () => {
             expect(aTeam.isInPlay(i)).toBeTruthy()
         }
     })
+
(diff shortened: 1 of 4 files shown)


app/models/Team.tsx
```diff
@@ -0,0 +1,32 @@
...
+    teamName: string
+    relationship: string
+    isParticipating: boolean
+    eliminationOrder: number
+
+    constructor(inTeam: ITeam) {
+
+        if ((inTeam.eliminationOrder === 0 && !inTeam.isParticipating) ||
+            (inTeam.eliminationOrder !== 0 && inTeam.isParticipating)) {
+            console.warn("Building a team whose eliminationOrder is 0 (default), but they are also have isParticipating = false. May be a bad team construction")
```
CAnthonySiever, line 13: We might want to include something to identify this possibly invalid team in the warning.

__tests__/models/Team.js
```diff
@@ -0,0 +1,79 @@
...
+                eliminationOrder: i
+            })
+
+            // Act, Assert
+            expect(aTeam.eliminationOrder).toBe(i)
+            expect(aTeam.isInPlay(i-1)).toBeFalsy()
+        }
+    })
+
+    it('should be true if roundOrder is two less than eliminationOrder', () => {
```
CAnthonySiever, line 64: I'm a bit confused on why this would be the case, but it seems very intentional so I trust it 😅 Would love an explanation, but that can wait.

CAnthonySiever (approved):

app/models/Team.tsx
jjm3x3: That's a good suggestion. Any idea how to do that? Do you think we can rely on `teamName` here?

__tests__/models/Team.js
jjm3x3: So I just added a comment that I hope clears this up, but the TL;DR is because we are dealing with an off by one error. I expect that you might expect to see something like this around values -1,0 and/or 1, but the 2 is probably what threw you off, right?

As an aside, I do think this could be fixed if we updated the `eliminationOrder` property to also be 0 indexed. Certainly worth considering, though even that is a bit confusing at first glance as well 😅

app/models/Team.tsx
CAnthonySiever: `teamName` might give more information if something were going wrong, but this might warrant an additional `id` field that would be unique for each Team object, offering insights into potential issues with duplicate teams, incomplete teams, and empty teams.

CAnthonySiever (approved):

app/models/Team.tsx
jjm3x3: Exactly! I would go so far as to say that this seems fairly intertwined with #25

PR #30, opened 2024-01-13
Feature/cross out based on list position

jjm3x3:
In this PR we see how the rest of my plan starts to come together. The main feature of this PR is I have introduced a helper method which will take in a list of teams a team and a round number and determine if that team should be scored or crossed out. At first glance those two may not seem like clear opposites and you might be right, but for all intents and purposes we are going to treat them that way for now. I did find out that given how we are expecting to score this and display it tend to coincide quite nicely and for now I would like to take advantage of that.

Few things to note, in no particular order
- Could this util have gone in a class - maybe, but I decided that was overkill for now
- Round is assumed to be 0 indexed because why not all other programing things are
- `teamPosition` is 1 indexed. This could probably could change, but it's really just a side effect of how it is calculated (see next point)
- `teamPosition` is the position that the team is in starting from the right most position, this is because we want to display the list with the crossed off eliminated teams at the bottom of the list
- The name of the function is `shouldBeScored` but it is also being used to cross off and that is also this branch/PR name. Yes all that is true see PR description and (see previous PR #29 ) where I am trying to merge these two things. I think this is a decent benefit for now to have this logic merged since we then get some other behaviors for free, but I could be convinced otherwise

### DIFF
--- a/__tests__/utils/teamListUtils.js
+++ b/__tests__/utils/teamListUtils.js
@@ -15,7 +15,11 @@ describe('teamListUtils shouldBeScored', () => {
         expect(result).toBeFalsy()
     })
 
-    it('should be false when the target team is in the first position or end of array and its the first round', () => {
+    it('should be false when the target team is at the end of array and its the first round', () => {
+        // Note: being at the end of the array assums that that team should he
+        //   eliminated first
+        // Note2: rounds are 0 indexed so the first round is 0
+
         // Arrange
         const aTeam = {
             isInPlay: (round) => true
@@ -29,7 +33,8 @@ describe('teamListUtils shouldBeScored', () => {
         expect(result).toBeFalsy()
     })
 
-    it('should be false when the target team is in the first position or end of array and its the second round round', () => {
+    it('should be false when the target team is at the end of array and its the second round round', () => {
+
         // Arrange
         const aTeam = {
             isInPlay: (round) => true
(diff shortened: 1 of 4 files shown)


internette (approved):

__tests__/utils/teamListUtils.js
```diff
@@ -0,0 +1,59 @@
...
+        // Assert
+        expect(result).toBeFalsy()
+    })
+
+    it('should be false when the target team is in the first position or end of array and its the first round', () => {
+        // Arrange
+        const aTeam = {
+            isInPlay: (round) => true
+        }
+        const teamList = [{}, aTeam]
```
CAnthonySiever, line 23: So I see how it's at the end of the array but I'm a bit confused as to how this is testing it for both the first position and the end of the array. Perhaps this should be made into two tests.

__tests__/utils/teamListUtils.js
```diff
@@ -0,0 +1,59 @@
...
+        // Assert
+        expect(result).toBeFalsy()
+    })
+
+    it('should be false when the target team is in the first position or end of array and its the second round round', () => {
+        // Arrange
+        const aTeam = {
+            isInPlay: (round) => true
+        }
+        const teamList = [{}, aTeam]
```
CAnthonySiever, line 37: Same question as above.

CAnthonySiever (approved):

__tests__/utils/teamListUtils.js
jjm3x3: Updated the test name and added a comment we can reopen this or create an issue if you feel strongly about it!

__tests__/utils/teamListUtils.js
jjm3x3: Updated the test name. See previous reply: [here](https://github.com/jjm3x3/AmazingRaceFantasy/pull/30#discussion_r1461430830)

PR #31, opened 2024-01-14
Feature/parse out final elimination orders

jjm3x3:
Here is a small thing that I only realized after the season ended. The way the Wiki article is written is the teams in the finale don't get an elimination order the same way that the other teams do instead they are labeled 'Third place' and 'Runners-up'. So this will fix that to give them an elimination order so we can score the order that these teams come in.


CAnthonySiever (approved):

CAnthonySiever:
Looks good! No notes.

PR #33, opened 2024-01-23
Feature/create round list component

jjm3x3:
Building a bit off the previous concept it felt a bit easier to create a component for the list of rounds to pass around in a future change, also it made writing these out since we can now have the roundList component capture the grand total score entirely. The only other thing I could imagine, that might make this a bit better is considering a world where we let the component generate the round scores also, but I figure this could easily come after some refactor of the page itself.

### DIFF
--- a/app/components/contestantRoundList.tsx
+++ b/app/components/contestantRoundList.tsx
@@ -19,7 +19,7 @@ export default function ContestantRoundList({
     return (
         <div className="text-center">
             {perfectRoundScores.map((perfectScore, roundNumber) => {
-                grandTotal += score
+                grandTotal += perfectScore
                 let contestantRoundScore = contestantRoundScores[roundNumber]
                 contestantGrandTotal += contestantRoundScore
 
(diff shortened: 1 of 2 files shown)


app/components/contestantRoundList.tsx
```diff
@@ -0,0 +1,39 @@
...
+                grandTotal += score
+                let contestantRoundScore = contestantRoundScores[roundNumber]
+                contestantGrandTotal += contestantRoundScore
+
+                return <Round
+                    key={"round"+roundNumber}
+                    roundNumber={roundNumber}
+                    perfectTeamList={perfectTeamList}
+                    contestantTeamList={contestantTeamList}
+                    perfectWeekScore={score}
```
CAnthonySiever, line 31:
```suggestion
                    perfectWeekScore={perfectScore}
```

Purely a suggestion but score certainly works here.

CAnthonySiever (approved):

app/components/contestantRoundList.tsx
jjm3x3: I'm not going to accept this specific change only because it should be made intandum with renaming/declaring the variable so I will add a whole other comit.

PR #34, opened 2024-01-27
Feature/extract score generation

jjm3x3:
Here is just a few commits to make sure we are getting some sensible code reuse. This same scoring function will be used over and over again and so it might as well be extracted into a proper function. This does seem like a pretty good candidate to be tested so before this gets merged we should look into adding testing

## TODO
- [ ] Add testing

### DIFF
--- a/app/scoring/page.tsx
+++ b/app/scoring/page.tsx
@@ -64,17 +64,7 @@ export default async function Scoring() {
         return foundTeam
     })
 
-    const contestantRoundScores: number[] = []
-    for(let i = 0; i < numberOfRounds; i++) {
-        const roundScore = currentSelectedContestantTeamsList.reduce(
-            (acc: number, x: Team) => {
-                const teamShouldBeScored = shouldBeScored(currentSelectedContestantTeamsList, x, i)
-
-                return teamShouldBeScored ? acc + 10 : acc
-            }, 0)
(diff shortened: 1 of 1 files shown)


internette (approved):

jjm3x3:
Given this is really just undoing a copy paste job, I am thinking I wont bother adding testing for this right now. This is mostly just a small internal refactor for some other changes yet to come.

app/scoring/page.tsx
```diff
@@ -8,6 +8,24 @@ interface Dictionary<T> {
...
 
+
+function generateContestantRoundScores(contestantTeamsList: Team[], numberOfRounds: number) {
+
+    const contestantRoundScores: number[] = []
+
+    for(let i = 0; i < numberOfRounds; i++) {
+        const roundScore = contestantTeamsList.reduce(
+            (acc: number, x: Team) => {
+                const teamShouldBeScored = shouldBeScored(contestantTeamsList, x, i)
```
CAnthonySiever, line 19: For some reason Github suggestions aren't available here 😅 but I'm thinking something like `team` or `contestantTeam` instead of `x` here.

CAnthonySiever: Ah, it's already merged 😅

PR #46, opened 2024-01-31
Enforce node version

internette:
This ensures that we are all developing against the same version of node

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+engine-strict=true
--- a/package.json
+++ b/package.json
@@ -9,6 +9,9 @@
     "lint": "next lint",
     "test": "jest --watch"
   },
+  "engines": {
+    "node": ">=20.11.0"
+  },
   "dependencies": {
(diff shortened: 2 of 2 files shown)


.npmrc
```diff
@@ -0,0 +1 @@
+engine-strict=true
```
jjm3x3, line 1: We like newlines at EOF 😁 [source](https://stackoverflow.com/a/729795)

jjm3x3 (approved):
Not sure if you can merge, but worth a shot. Also small nit if you don't mind.

package.json
```diff
@@ -9,6 +9,9 @@
     "lint": "next lint",
     "test": "jest --watch"
   },
+  "engines": {
+    "node": ">=20.11.0"
```
jjm3x3, line 13: I am not sure we should use `>20.x` since this is still just in beta for Vercel
image

jjm3x3 (changes requested):

jjm3x3 (approved):
LGTM

PR #48, opened 2024-02-03
Feature/add gh action

jjm3x3:
Based with some of the difficulty in #35 implementing testing, it became clear that this should be done through means other than the Vercel build command. So this PR is to setup a github action that will run the build and test so that we have complete control over the testing command.


CAnthonySiever (commented):
Looks good! 👍

internette (approved):

Issue #42, opened 2024-01-29
Add ability to pull from a different Wikipedia source


jjm3x3:
Based on how the site was initially build the one location that the `utils/wikiFech` util could pull from was hard coded into the util. This should be updated to allow different components and pages to bring their own for different leagues.

PR #53, opened 2024-02-04
Feature/select wiki source

jjm3x3:
This should handle #42 context in there!

### DIFF
--- a/app/contestants/page.tsx
+++ b/app/contestants/page.tsx
@@ -3,7 +3,7 @@ import { wikiUrl, apiUrl, getWikipediaContestantData } from "../utils/wikiFetch"
 
 export default async function Contestants() {
 
-    const wikiContestants = await getWikipediaContestantData()
+    const wikiContestants = await getWikipediaContestantData(apiUrl)
     const final = getTeamList(wikiContestants)
 
     return (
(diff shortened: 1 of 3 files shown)


app/contestants/page.tsx
```diff
@@ -1,9 +1,9 @@
 import { getTeamList, ITeam } from "../utils/wikiQuery"
-import { wikiUrl, getWikipediaContestantData } from "../utils/wikiFetch"
+import { wikiUrl, apiUrl, getWikipediaContestantData } from "../utils/wikiFetch"
 
 export default async function Contestants() {
 
-    const wikiContestants = await getWikipediaContestantData()
+    const wikiContestants = await getWikipediaContestantData(apiUrl)
     const final = getTeamList(wikiContestants)
```
internette, line 7: You could convert these two lines into a util function called something like "getPageData"

app/scoring/page.tsx
```diff
@@ -29,7 +29,7 @@ function generateContestantRoundScores(contestantTeamsList: Team[], numberOfRoun
 
 export default async function Scoring() {
 
-    const wikiContestants = await getWikipediaContestantData()
+    const wikiContestants = await getWikipediaContestantData(apiUrl)
     const pageData = getTeamList(wikiContestants)
```
internette, line 33: And reuse the util here

internette (commented):
Needs unit test

app/contestants/page.tsx
jjm3x3: This is a decent idea, but I am not yet convinced that now is the right time to do that. [see Rule Of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))

app/scoring/page.tsx
jjm3x3: Same as above.

jjm3x3:
> Needs unit test

What should I unit test here?

internette:
👍L(almost)GTM👍

Issue #37, opened 2024-01-28
Displaying Firendly Team Names


jjm3x3:
What was called out in a bit of customer feedback is that the currently displayed team names (with each team members full name, first and last), is a bit difficult to read through and unnecessary. The teams were originally shown this way since this has been the only reliable way to pull team names from Wikipedia and tie those teams to the ones listed by a league contestant's team list ranking. While that is still an issue that may need to be addressed (see #25) we can certainly attempt to move towards only displaying team members first name as part of the team name since that will be a bit more readable.

jjm3x3:
As I have thought more about this issue there is still the issue of figuring out when a person has two first names or two last names or some combination of the two and at the time of writing this the only thing I can think of is add a list of well known name pairs to look up to see if any given name should be considered as just one name or two.

jjm3x3:
I also noticed that one of the tables in the Wikipedia article for Amazing Race 35 has a list of team names where this is already figured out. I wonder if we might be able to add in a shortcut using that. FFT

PR #54, opened 2024-02-04
Feature/friendly team display name

jjm3x3:
Based on some customer feedback, displaying team contestants full name is a bit much, especially on mobile. So here is a fairly simple fix in order to clean that up!

This is an attempt at #37


internette (approved):
And testing too? Happy birthday month to meeeeee

Issue #55, opened 2024-02-08
Extract Contestant Ranking


jjm3x3:
In attempt to minimize the Scoring page and to start taking data a bit further out of the code. It makes sense to extract the Contestant Ranking into dedicated files which can hold that data. This could also be used as a place to start considering what kind of data structure could benefit us going forward.

PR #58, opened 2024-02-11
Feature/extract league data

jjm3x3:
This is a pretty simple PR which does exactly what it says and accomplishes #55


CAnthonySiever (approved):
Apologies on the delay; finally got around to reading this 😅 No direct notes on this. 

I am wondering if it might be best to have a form at the start of each season where each contestants' rankings can be filled out, then a json file for the season (or if it turns out to be easier, a json file for each contestant for each season) is generated.

jjm3x3:
> Apologies on the delay; finally got around to reading this 😅 No direct notes on this.
> 
> I am wondering if it might be best to have a form at the start of each season where each contestants' rankings can be filled out, then a json file for the season (or if it turns out to be easier, a json file for each contestant for each season) is generated.

Completely agree and I think the first step to this is actually the refactor I did in #59. I don't have an issue dedicated to that, but I realized that #34 would be a lot easier once I have taken care of that!

Issue #41, opened 2024-01-29
Extract Scoring Content Generation


jjm3x3:
Since each season for now will have the same scoring generation that will be done during build time or at least be Server Side rendered we will need to make sure that there is a fairly simply way to spin up new scoring pages across new leagues/seasons.

PR #61, opened 2024-02-21
Feature/generate contestant data

jjm3x3:
There isn't necessarily a lot going on in this PR, but I do feel like this concept is a big one since it will likely be how lots of other pages work, especially ones that have a lot of static content and need to be generated for every league. The idea here is a simple one. We want to have at least one scoring page for every league but the contestants and the Wikipedia data backing the scoring will change for each league. In order to re use as much of the code as possible I am extracting a lot of it here into a "generator" that way that same generator can be invoked in another scoring page for another league/season.

This should address #41

### DIFF
--- a/app/scoring/page.tsx
+++ b/app/scoring/page.tsx
@@ -8,28 +8,6 @@ import { CONTESTANT_LEAGUE_DATA } from '../leagueData/AmazingRace_35'
 import ContestantSelector from '../components/contestantSelector'
 import generateListOfContestantRoundLists from '../generators/contestantRoundListGenerator'
 
-interface Dictionary<T> {
-    [Key: string]: T;
-}
-
-
-function generateContestantRoundScores(contestantTeamsList: Team[], numberOfRounds: number) {
-
-    const contestantRoundScores: number[] = []
-
-    for(let i = 0; i < numberOfRounds; i++) {
-        const roundScore = contestantTeamsList.reduce(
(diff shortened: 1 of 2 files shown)


app/scoring/page.tsx
```diff
@@ -1,70 +1,11 @@
...
-                const teamShouldBeScored = shouldBeScored(contestantTeamsList, x, i)
-
-                return teamShouldBeScored ? acc + 10 : acc
-            }, 0)
-        contestantRoundScores.push(roundScore)
-    }
-
-    return contestantRoundScores
-}
+import generateListOfContestantRoundLists from '../generators/contestantRoundListGenerator'
```
internette, line 4: Yayyyyyyy

app/scoring/page.tsx
```diff
@@ -1,70 +1,11 @@
...
-            return foundTeam
-        })
-
-        const contestantsRoundScores: number[] = generateContestantRoundScores(contestantsTeamList, numberOfRounds)
-
-        return {
-            key: contestant.name,
-            content: <ContestantRoundList perfectRoundScores={roundScores} contestantRoundScores={contestantsRoundScores} perfectTeamList={reverseTeamsList} contestantTeamList={contestantsTeamList}/>
-        }
-    })
```
internette, line 67: More yayyyyyy!

internette (approved):

Issue #60, opened 2024-02-18
Swap old pages and new


jjm3x3:
This is very much more along the lines of a task since it is fairly mechanical. The idea here is because of how the site is organized it can only support one active league at a time. Since that meets our needs for now we can leave that aspect of the site as is. However since a new season and a new league is also starting as a result, we would like the "active" pages (ie the ones that are linked to from the home page) to be the ones for the new league. This doesn't need to be done until some time in the week leading up to the new season starting, but once it is done the pages for the new league should be linked from the home and the pages for the old league can still be available, but are hosed at some deeper path (url).

jjm3x3:
One suggestion is `/archive/AmazingRace_36` though IDK how the capital letters and the underscore will work with the URL pathing. The idea here is to possibly match the files hosed in the `/leagueData` and `/leagueConfiguration` directories.

PR #63, opened 2024-03-02
Feature/swap old new pages

jjm3x3:
This is the final code change necessary to swap the pages in for the new season of the Amazing Race and the league that will be built for it. In this change I am taking the advice described in #60 . So the old pages will be put in a location which won't collide with any other old leagues in the `/archive/AmazingRace_35/` dir/path so that it is simple to get back to them, but they should never need to move again.


CAnthonySiever (approved):
Everything looks good! 👍 Are we planning on keeping an archive for each season like this?

jjm3x3:
> Everything looks good! 👍 Are we planning on keeping an archive for each season like this?

Yeah, I am thinking it's easy enough to do, plus people might want to go back and see what happened in previous leagues. WDYT?

Also keep in mind that in the near future I am thinking of doing a redesign that will expose these pages in the UI.

CAnthonySiever:
> > Everything looks good! 👍 Are we planning on keeping an archive for each season like this?
> 
> Yeah, I am thinking it's easy enough to do, plus people might want to go back and see what happened in previous leagues. WDYT?
> 
> Also keep in mind that in the near future I am thinking of doing a redesign that will expose these pages in the UI.

Yeah it makes sense to have something like "View previous winners or scores" or something along those lines.

internette:
Fix your goddamn url naming

jjm3x3:
> Fix your goddamn url naming
> 
> Edit for professional sake: Please fix your url naming

I am going to need some more actionable feedback in order to address this... 😞

internette:
> > Fix your goddamn url naming
> > Edit for professional sake: Please fix your url naming
> 
> I am going to need some more actionable feedback in order to address this... 😞

All URLs should be 100% lowercase. If you want to separate words, preferably use underscores. You *can* use dashes pr %20, but that's ill-advised.

jjm3x3:
> > > Fix your goddamn url naming
> > > Edit for professional sake: Please fix your url naming
> > 
> > 
> > I am going to need some more actionable feedback in order to address this... 😞
> 
> All URLs should be 100% lowercase. If you want to separate words, preferably use underscores. You _can_ use dashes pr %20, but that's ill-advised.

-----

> All URLs should be 100% lowercase.

I would like some source to back this up

> If you want to separate words, preferably use underscores. You _can_ use dashes pr %20, but that's ill-advised.

Also a quick web search and [one of the top results, from google](https://developers.google.com/search/docs/crawling-indexing/url-structure) actually suggests `-` over `_`? What's that about?

internette (approved):

PR #67, opened 2024-03-03
Use anchor to find the current section number to avoid pulling bad data

CAnthonySiever:
As the title implies, this adds another moving part to the pulling of data, but it seems warranted since the section number from Wikipedia that we're looking to pull from is subject to change. The anchor seems to be a bit more stable and even consistent across different seasons.

I would suggest playing with it a bit since it's fun to go around the different seasons just by changing the 35 in `WIKI_API_URL` to 3 and so on in a dev environment as this might inspire other changes and features elsewhere 😁

Edit: I guess this also needs a reference to this issue since it is the final part of resolving it #5 😅

### DIFF
--- a/app/utils/wikiFetch.tsx
+++ b/app/utils/wikiFetch.tsx
@@ -14,6 +14,25 @@ export interface IWikipediaContestantData {
     status: string
 }
 
+interface Section {
+    toclevel: number
+    level: string
+    line: string
+    number: string
+    index: string
+    fromtitle: string
+    byteoffset: number
+    anchor: string
+    linkAnchor: string
+}
+
+interface ParseResult {
+    title: string
+    pageid: number
+    sections: Section[]
+    showtoc: boolean
+}
+
 async function fetchWikipediaData(wikiUrl: string): Promise<IWikipediaData> {
     const response = await fetch(wikiUrl, { next: { revalidate: 3600 } })
     const data = await response.json()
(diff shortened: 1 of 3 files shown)


internette (approved):

app/utils/wikiFetch.tsx
```diff
@@ -14,15 +14,56 @@ export interface IWikipediaContestantData {
...
 async function fetchWikipediaData(wikiUrl: string): Promise<IWikipediaData> {
     const response = await fetch(wikiUrl)
     const data = await response.json()
     return data
 }
 
+export async function fetchWikipediaSections(wikiUrl: string): Promise<ParseResult> {
+    const response = await fetch(wikiUrl)
+    const data = await response.json()
+    return data.parse
```
jjm3x3, line 45: Curious, should we do something similar in `fetchWikipedidiaData`? Doesn't need to happen in this PR, just wondering on thoughts 🤔

jjm3x3 (approved):
Just some food for thought. I know it can be a bit difficult to get unit tests in here since this connector is meant do integration so there is no simple way to stub out the call, however it would be cool if we could run an integration test or two just so that we can get some signal if this logic stops working at some point. Otherwise, great work! 🙏

jjm3x3:
Also _pro tip:_ if you want the issue to auto close w/ the PR you need to attach it in the "Development" section of the PR: 
image

app/utils/wikiFetch.tsx
internette: Definitely, if we plan to keep using Wikipedia for multiple shows

internette (approved):

PR #74, opened 2024-03-16
Feature/link to sheet

### LINKED ISSUE #73: Link to Google Sheets
So far both of the leagues that we have run have add a lot of tie in w/ Google Sheets. The first one was run almost entirely from Google Sheets and the second one (this one) has a Google form that is feeding all it's results into a Google Sheet. Then also Andrew has agreed to keep score parallelly in that sheet as this will be the sites first time trying to keep score. (gotta keep the robots honest)! So the goal here is to add links to the google sheet(s) somewhere in the site.

jjm3x3:
Would love some feedback on this to see what folks think. The way I figure it, it is ok to expose these sheets as long as they are view only to the public. Also it seems to me this is the right place, w/ the right messaging, but would love to hear what others think!

Closes #73

### DIFF
--- a/app/scoring/page.tsx
+++ b/app/scoring/page.tsx
@@ -15,7 +15,7 @@ export default async function Scoring() {
             <h1 className="text-2xl text-center">Current Scoring for</h1>
             <br/>
             <ContestantSelector listOfContestantRoundLists={listOfContestantRoundLists}/>
-            <p>This seaons contestant data has been persisted into <a className="standard-link" href={GOOGLE_SHEET_URL}>this google sheet</a> from a google form.</p>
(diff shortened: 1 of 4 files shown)


app/scoring/page.tsx
```diff
@@ -15,6 +15,8 @@ export default async function Scoring() {
             <h1 className="text-2xl text-center">Current Scoring for</h1>
             <br/>
             <ContestantSelector listOfContestantRoundLists={listOfContestantRoundLists}/>
+            <p>This seaons contestant data has been sourced from <a className="standard-link" href={GOOGLE_SHEET_URL}>this google sheet</a> which was populdated using a google form.</p>
```
internette, line 18: Nit:
populated is misspelled

internette (approved):

Issue #64, opened 2024-03-02
Create a League Standing Page


jjm3x3:
In order to see how you or any one contestant in the league is doing in comparison to any other contestant, we would like to present a league standing page so that it can be seen at a glance how all the league contestants are fairing against one another. Once complete there should be a page which will show the current standing for all contestants in the league including their total scores for the week.

PR #80, opened 2024-03-30
Feature/add league standing page

jjm3x3:
Here is the final part for #64... more details TBD


internette (approved):

internette (commented):
Update page path name

Issue #86, opened 2024-04-14
Add a Hamburger Menu


jjm3x3:
The ask is to build a hamburger menu which will show in the upper right corner. Once open it should be a scroll-able list with an "X" replacing where the menu used to be in order to close it. The top of the list should be an option to go home.

PR #94, opened 2024-04-29
Hamburger Menu Implementation

internette:
Created hamburger menu and uses the data from the homepage to fill the navigation.

image
image
image


jjm3x3:
Friendly suggestion on "process" can you make sure this links to the corresponding issue #86 that way when this gets merged it will update the status of that issue, also the pages can blinked together. Also I would say this PR is a good use-case for the "test: UAT" label.

Quick UAT related thing, I see that when one is viewing it from a smaller view port the drop-down can cover up the main content like in this screenshot
image

app/components/navigation/navigation-item.tsx
```diff
@@ -0,0 +1,10 @@
+import { useState } from "react";
+import INavigationItem from '@/app/models/INavigationItem';
+export default function NavigationItem({ inputAttr, labelAttr, listAttr, children, hasSubpages }: INavigationItem) {
+    const [isHidden, setIsHidden] = useState(true);
+    return (<>
+        <input id={inputAttr.id} className={inputAttr.classes} type="checkbox" onChange={(e)=> setIsHidden(!e.target.value)}/>
```
jjm3x3, line 6: nit: I would say formatting this so that it can be more friendly to read on narrow screens can be a nice optimization in code quality and review. It can also make diffs easier to read.
```suggestion
        <input
            id={inputAttr.id}
            className={inputAttr.classes}
            type="checkbox"
            onChange={(e)=> setIsHidden(!e.target.value)}
        />
```

app/components/navigation/navigation-item.tsx
```diff
@@ -0,0 +1,10 @@
...
+        <label htmlFor={inputAttr.id} id={labelAttr.id} aria-controls={labelAttr.aria.controls} data-testid={labelAttr.testId} className={labelAttr.classes }>{labelAttr.content}</label>
```
jjm3x3, line 7: nit: same as above, though, I am less sure how to fix this one since the tag has content 😅

app/components/navigation/navigation-item.tsx
```diff
@@ -0,0 +1,10 @@
...
+        {hasSubpages && <ul className={listAttr.classes}  id={listAttr.id} data-testid={listAttr.testId} aria-hidden={isHidden} hidden={isHidden}>{children}</ul>}
+    </>)
+}
```
jjm3x3, line 10: nit: do we not expect newlines EOF yet 😅

jjm3x3 (commented):
Just a few comments, haven't made it very far 😅

app/components/navigation/navigation-item.tsx
internette: > nit: I would say formatting this so that it can be more friendly to read on narrow screens can be a nice optimization in code quality and review. It can also make diffs easier to read.

Agreed. I'll make this change here. I've also created an issue to tackle stuff like this with auto-formatting: https://github.com/jjm3x3/AmazingRaceFantasy/issues/97

app/components/navigation/navigation-item.tsx
internette: Haha fair. I'm used to this being automatically added, so I'm still adjusting to manually adding it.

jest.config.mjs
```diff
@@ -11,6 +11,8 @@ const config = {
   // setupFilesAfterEnv: ['<rootDir>/jest.setup.js'],
 
   testEnvironment: "jest-environment-jsdom",
+  setupFilesAfterEnv: ["<rootDir>/__tests__/setupTests.js"],
+  testPathIgnorePatterns: ["<rootDir>/build/", "<rootDir>/node_modules/", "<rootDir>/__tests__/setupTests.js"]
```
jjm3x3, line 15: Interested to know how you arrived at these values. the `setupTests.js` seems fairly obvious, but the other ones are a bit interesting. Are those defaults that now need to be specified since we are including this config, or are those ones that are usually added and just where not added originally because I set it up and still learning 😅

app/page.tsx
```diff
@@ -1,49 +1,33 @@
...
-    const currentDirFilesList = fs.readdirSync(__dirname)
-    let archiveDirFilesList: string[] = []
-
-    if (currentDirFilesList.includes("archive")) {
-        console.log("Yay there is an 'archive' folder and it's contents are:")
-        archiveDirFilesList = fs.readdirSync(__dirname+"/archive")
-        console.log(archiveDirFilesList)
-    }
-
+    const pages = getPages() || [];
```
jjm3x3, line 8: fft: this seems like a simple enough refactor that could have been pulled into a separate PR and not get wrapped up in all the bigger more important changes you are doing. 😅

jjm3x3 (commented):
Just another round of comments, nearing half way through.

app/components/navigation/navigation-item.tsx
```diff
@@ -0,0 +1,20 @@
...
+            type="checkbox"
+            onChange={(e)=> setIsHidden(!e.target.value)}
+        />
+        <label 
+            htmlFor={inputAttr.id}
+            id={labelAttr.id}
+            aria-controls={labelAttr.aria.controls}
+            data-testid={labelAttr.testId}
+            className={labelAttr.classes }>{labelAttr.content}</label>
+        {hasSubpages && <ul className={listAttr.classes}  id={listAttr.id} data-testid={listAttr.testId} aria-hidden={isHidden} hidden={isHidden}>{children}</ul>}
```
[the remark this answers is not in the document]
internette: This will be fixed in the auto-format ticket :)

app/components/navigation/navigation.module.scss
jjm3x3, line 1: Not a blocker, but I have been trying to convert as many things to tailwind as I reasonably can. Might be worth checking out for consistency.

jjm3x3 (commented):
Very close, just looking over the sass/css that we added for the nav...

app/components/navigation/navigation.module.scss
```diff
@@ -0,0 +1,70 @@
...
+        right: 0.5rem;
+    }
+}
+
+.sub-level-link {
+    padding: 0.25rem 0.5rem;
+    display: inline-block;
+    width: 100%;
+    &:hover {
+        cursor: pointer;
```
jjm3x3, line 37: I actually think this is unnecessary, right?
```suggestion
```

jjm3x3 (approved):
I think this is looking good, just lots of questions and nit picks, but this does what we need it to do and quite well, Nice Job!

app/models/INavigationItem.tsx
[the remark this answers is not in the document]
internette: I'll leave it as is right now. If we decide to change, we can refactor then

app/components/navigation/navigation.tsx
```diff
@@ -0,0 +1,61 @@
...
+                id: `subpage-${keyName}-dropdown`,
+                testId: `subpage-${keyName}-dropdown`,
+                classes: styles['nav-subpages-list']
+            }
+            return (<li key={`nav-toplevellink-${keyName}`} className={styles['nav-toplevel-page']}>
+                <NavigationItem hasSubpages={hasSubpages }  inputAttr={subpageInputAttr } labelAttr={subpageLabelAttr } listAttr={ subpageListAttr} children={page.subpages.map((subpage: ISubpage) => {
+                        const subpageKeyName = subpage.name.toLowerCase().replaceAll(' ', '-');
+                        return <li key={`nav-toplevellink-${keyName}-sublink-${subpageKeyName}`}>
+                            <Link href={subpage.path} className={styles['sub-level-link']}>{subpage.name}</Link>
+                        </li>
```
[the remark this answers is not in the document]
internette: I'll create a separate PR for this

PR #99, opened 2024-05-10
97 Update Eslint

internette:
Updated ESLint and created a git hook that enables autoformatting on git commit, focused on the following file formats:
- ts
- tsx
- js
- jsx
- css
- scss

Background:
The .git file cannot be committed, so we needed to create a new place to store custom hooks. The approaches:
1. Add a npm package like [Husky](https://www.npmjs.com/package/husky) or [Pre-commit](https://www.npmjs.com/package/pre-commit)
2. Write a custom script to change the git hooks path

The best option seemed like a custom bash script. The script identifies if the engineer has Git version >=2.9 because the simplest command requires such. After the check, the script runs the appropriate command to set git's hook location local to this project.

I've also created an npm script in package.json for `setup:gitHooks` that makes the hooks executable and runs that script mentioned above. I've also added a `setup` command to set up the hooks and run `npm install` as a basic setup for this repo.

In this process, I've also migrated the ESLint config, since our previous implementation is now deprecated.


.githooks/pre-commit
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
```
jjm3x3, line 1: How does this work on windows do you know, else I can RTFM 🤔 👀

internette: No idea. I haven't checked on my PC

.github/workflows/checkSetupGithubHooksScript.yml
jjm3x3, line 1: Let's remove commented out code.

package.json
```diff
@@ -8,19 +8,19 @@
...
     "lint": "next lint",
     "test": "jest",
-    "test:watch": "jest --watch"
+    "test:watch": "jest --watch",
+    "format": "./node_modules/.bin/eslint -c 'eslint.config.mjs' .",
+    "format:fix": "./node_modules/.bin/eslint -c 'eslint.config.mjs' --fix ."
   },
   "engines": {
-    "node": ">=18.0.0"
+    "node":"^18.18.0 || ^20.9.0 || >=21.1.0"
```
jjm3x3, line 16: How come we switch the node 18 dep to use the `^` as opposed to the `>=` qualifier?

jjm3x3 (commented):
Only technically reviewed the first ~2000 lines of the `package-lock.json` it's so big 😧 ! Another thought comes to mind about how standard all this is because I am not familiar with the JS world. Definitely a step in the right direction though!

package.json
internette: Because the upgrade is only compatible with version 18.18, 20.9, or anything greater than version 21

internette: We discussed and we're using ^20.9.0

jjm3x3: Based on our new understanding the `^` means all minor and patch versions stopping at the next major version. Given that, and the fact that we don't really probably need to be on node 18 any more and more over we should be careful relying on odd versions of node since they seem to have a very short lifecycle. [node versions](https://nodejs.org/en/about/previous-releases)

internette:
> Only technically reviewed the first ~2000 lines of the `package-lock.json` it's so big 😧 ! Another thought comes to mind about how standard all this is because I am not familiar with the JS world. Definitely a step in the right direction though!

package-lock.json
```diff
@@ -2779,20 +3430,105 @@
       "integrity": "sha512-iKuQcq+NdHqlAcwUY0o/HL69XQrUaQdMjmStJ8JFmUaiiQErlhrmuigkg/CU4E2J0IyUKUrMAgl36TvN67MqTw==",
       "dev": true
     },
+    "node_modules/cssstyle": {
```
jjm3x3, line 3433: Dropping a checkpoint for myself 😅

jjm3x3 (commented):
So far it's looking fine, just working my way through `package-lock.json` still

package-lock.json
```diff
@@ -5975,14 +7142,12 @@
       }
     },
     "node_modules/lru-cache": {
-      "version": "6.0.0",
-      "resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-6.0.0.tgz",
-      "integrity": "sha512-Jo6dJ04CmSjuznwJSS3pUeWmd/H0ffTlkXXgwZi+eq1UCmqQwCh+eLsYOYCwY991i2Fah4h1BEMCx4qThGbsiA==",
+      "version": "5.1.1",
```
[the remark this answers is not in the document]
jjm3x3: Also upon a further look there is some weird stuff with this. I wonder if it does make sense to pin to 6.0.0 at least 😬

jjm3x3 (commented):
Just a few comments, no real blockers. About 2/3s of the way through `package-lock.json` 😮‍💨

package-lock.json
```diff
@@ -6554,10 +7760,34 @@
...
+        "minipass": "^5.0.0 || ^6.0.2 || ^7.0.0"
+      },
+      "engines": {
+        "node": ">=16 || 14 >=14.17"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/isaacs"
+      }
+    },
+    "node_modules/path-scurry/node_modules/lru-cache": {
```
jjm3x3, line 7778: I am just realizing the implication of this path. Does this mean that the resolved `lru-cache` for `path-scurry` is (or can be) different than the one resolved for everything else? 🤔

package-lock.json
```diff
@@ -802,21 +718,67 @@
...
+    "node_modules/@eslint/js": {
+      "version": "9.2.0",
+      "resolved": "https://registry.npmjs.org/@eslint/js/-/js-9.2.0.tgz",
+      "integrity": "sha512-ESiIudvhoYni+MdsI8oD7skpprZ89qKocwRM2KEvhhBJ9nl5MRh7BXU5GTod7Mdygq+AUl+QzId6iWJKR/wABA==",
+      "dev": true,
+      "engines": {
+        "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
       }
     },
     "node_modules/@humanwhocodes/config-array": {
```
jjm3x3, line 774: ~Funnily enough, this also seems like it's deprecated~ 😮 
_edit: I guess this is fine since this is really all because eslint is a whole major version behind_

package-lock.json
```diff
@@ -9,42 +9,39 @@
...
+        "sass": "^1.77.6",
         "tailwindcss": "3.3.3",
         "typescript": "5.2.2"
       },
       "devDependencies": {
+        "@eslint/js": "^9.2.0",
+        "@stylistic/eslint-plugin-js": "^2.1.0",
         "@testing-library/jest-dom": "^6.1.4",
         "@testing-library/react": "^14.1.2",
+        "eslint": "^8.57.0",
```
jjm3x3, line 29: I know I am to blame here as much as anyone, but I am just now seeing that version 9.x.x of eslint has been out for ~5mo. Is it worth at least putting an issue in this milestone, so that we can be kept up to date.

package-lock.json
```diff
@@ -8174,14 +9814,18 @@
       }
     },
     "node_modules/yallist": {
-      "version": "4.0.0",
-      "resolved": "https://registry.npmjs.org/yallist/-/yallist-4.0.0.tgz",
-      "integrity": "sha512-3wdGidZyq5PB084XLES5TpOSRA3wjXAlIWMhum2kRcv/41Sn2emQ0dycQW4uZXLejwKvg6EsvbdlVL+FYEct7A=="
+      "version": "3.1.1",
```
jjm3x3, line 9817: This looks like another downgrade that would be fixed by the LRU cache version pinning proposed in #120

jjm3x3 (commented):
Finally finished the review. I would say we should seriously consider pulling in #120 though that isn't necessarily a showstopper. WDYT?

package-lock.json
```diff
@@ -3589,20 +4403,59 @@
...
+    },
+    "node_modules/eslint/node_modules/@eslint/js": {
+      "version": "8.57.0",
+      "resolved": "https://registry.npmjs.org/@eslint/js/-/js-8.57.0.tgz",
+      "integrity": "sha512-Ys+3g2TaW7gADOJzPt83SJtCDhMjndcDMFVQ/Tj9iA1BfJzFKD9mAUXT3OenpuPHbI6P/myECxRJrofUsDx/5g==",
+      "dev": true,
+      "engines": {
+        "node": "^12.22.0 || ^14.17.0 || >=16.0.0"
+      }
+    },
```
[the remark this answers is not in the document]
internette: No idea

package-lock.json
internette: Yep

package-lock.json
internette: Yeah I think a follow up ticket should go in this milestone

package-lock.json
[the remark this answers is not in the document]
internette: Merged PR in

jjm3x3 (approved):
🚀

PR #105, opened 2024-05-13
Added default node version

internette:
- This autochanges to the minimum version of node if installed and nvm is available
- Updated the minimum version needed in package.json


jjm3x3 (approved):
LGTM

Issue #91, opened 2024-04-29
Tie subpages in with home redesign


jjm3x3:
Now that  the home page has been given the redesign treatment with #75 the sub pages should get a similar treatment along with getting the title bar and giving the users a way to navigate back to home.

jjm3x3:
As commented in #113 this has been closed as a result of that, but there is one follow on item being tracked with #114

PR #113, opened 2024-06-10
Feature/subpage redesign2

jjm3x3:
Given some of the issues we have encountered with putting the Navigation in the layout we are splitting up the work that was going to be shipped in #93. This is just some of the cosmetic work that was going to be updated for experience. And we will open a new issue to track moving the header onto all subpages #114.


internette (approved):

jjm3x3:
Assuming that merging this will close #91 adding that there is a follow on issue being tracked by #114

PR #115, opened 2024-06-10
Implementing core styles

internette:
Created a new style standard for our app:

- `_colors`: This is where we store our brand colors. We should limit our color uses to colors defined in this file.
- `_links`: This is where we keep our link styles.
- pagename.scss: We should create an individual scss file per page only when needed. Most values should exist in the component styles and global styles.
- index.scss: This is an index of all of our partial styles. This is imported into our `global.scss` file.
- `global.scss`: This was the `global.css` file.

TO DO:
Documentation around our new design system


app/styles/homepage.scss
```diff
@@ -0,0 +1,8 @@
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
+
+.standard-link-container {
+    padding: 15px;
+    @apply basis-1/3;
```
jjm3x3, line 7: Basis 1/3 doesn't work super well because of the text wrapping 😧 
image

app/globals.scss
```diff
@@ -1,19 +1,20 @@
+@use './styles/index.scss' as styles;
```
jjm3x3, line 1: Only needs colors now maybe we can go back to that.

jjm3x3 (commented):
Just a partial review so far, but here are a few things I noticed.

app/styles/homepage.scss
internette: Updated

image

app/page.tsx
```diff
@@ -22,9 +23,11 @@ export default function Home() {
                     const keyName = p.name.toLowerCase().replaceAll(' ', '-');
                     return (<div key={`links-section-${keyName}`}>
                     <p className="league-link-heading" >Links For { p.name } League</p>
-                    <div className="flex flex-row">{
+                    <div className="md:flex md:flex-row">{
                         p.subpages.map((pSub: ISubpage) => (
-                            <Link className="standard-link league-page-link" key={`links-section-${keyName}-link-${pSub.name.toLowerCase().replaceAll(' ', '-')}`} href={pSub.path}>{ pSub.name}</Link>
+                            <p className="standard-link-container md:basis-1/3">
+                                <Link className="standard-link league-page-link" key={`links-section-${keyName}-link-${pSub.name.toLowerCase().replaceAll(' ', '-')}`} href={pSub.path}>{ pSub.name}</Link>
```
jjm3x3, line 29: Don't think we need `league-page-link` anymore.
```suggestion
                                <Link className="standard-link" key={`links-section-${keyName}-link-${pSub.name.toLowerCase().replaceAll(' ', '-')}`} href={pSub.path}>{ pSub.name}</Link>
```

jjm3x3 (approved):
Just a small nit, otherwise looking good!

PR #116, opened 2024-06-10
Set up Pull Request Template

internette:


.github/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
+### Summary/Acceptance Criteria
+*Description of what this PR accomplishes*
+
+### Screenshots
+*If applicable*
+
+## Confirm
+- [ ] This PR has unit tests scenarios.
+- [ ] This PR is correctly linked to the relevant issue or milestone.
+- [ ] This PR has been locally QA'd.
```
jjm3x3, line 10: fft: perhaps we should also add an item to check on the deployed preview.

jjm3x3 (approved):
LGTM.

TBH I feel like this is something we could pick a part and try to make perfect, but I have never really used these templates before, so I just say *SEND IT* let's FAAFO 😁

PR #119, opened 2024-06-29
100 read in big brother data

jjm3x3:
Just as the corresponding issue suggests this will enable us to pull data from the BB wiki page. This does so by keeping most things pretty generic. We are giving up on the concrete model coming from Wikipedia and instead just returning a "row" object (with the exception of the name & name2 property). The other thing is we are going to start accepting the table name as an argument, this will help us scale beyond just the "Cast" table in Amazing Race, but will also allow us to use the "HouseGuest" table.

### Screenshots
N/A

## Confirm
- [x] This PR has unit tests scenarios. (no new ones necessary)

/assign me

### DIFF
--- a/__tests__/wikiQuery.js
+++ b/__tests__/wikiQuery.js
@@ -9,11 +9,11 @@ describe('getData', () => {
         const listOfContestants = [
             {
                 name: firstContestantsFirstName + " Guy",
-                status: "Participating"
+                col4: "Participating"
             },
             {
                 name: secondContestantsFirstName + " Brother",
-                status: "Participating"
+                col4: "Participating"
             }]
 
         // Act
@@ -31,7 +31,7 @@ describe('getData', () => {
         const listOfContestants = [
             {
                 name: firstContestantsFirstName + " Guy",
-                status: "Participating"
+                col4: "Participating"
             },
             {name: secondContestantsFirstName + " Brother"}
         ]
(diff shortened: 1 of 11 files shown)


CAnthonySiever (approved):
Looks good 👍 I think the [Cast table](https://en.wikipedia.org/wiki/The_Amazing_Race_25#Cast) from AR is pretty comparable to the [HouseGuests table](https://en.wikipedia.org/wiki/Big_Brother_25_(American_season)#HouseGuests) for BB, so while Col1-5 certainly works, and opens this up to other fantasy show scenarios, we could also probably get away with `name`, `age`, `details` (not sold on this but it's relationship for AR and Occupation for BB), `city`, `status`, or something like that. Also worth noting that BB seems to have a 6th column with `Entry`. Not entirely sure what this means or how important it is.

Approving this though since this is good, and puts the project in the right direction whether or not we consider any of the above.

internette (approved):

jjm3x3:
> Looks good 👍 I think the [Cast table](https://en.wikipedia.org/wiki/The_Amazing_Race_25#Cast) from AR is pretty comparable to the [HouseGuests table](https://en.wikipedia.org/wiki/Big_Brother_25_(American_season)#HouseGuests) for BB, so while Col1-5 certainly works, and opens this up to other fantasy show scenarios, we could also probably get away with `name`, `age`, `details` (not sold on this but it's relationship for AR and Occupation for BB), `city`, `status`, or something like that. Also worth noting that BB seems to have a 6th column with `Entry`. Not entirely sure what this means or how important it is.
> 
> Approving this though since this is good, and puts the project in the right direction whether or not we consider any of the above.

I do tend to agree, however I was leaning towards the more generic fields for now because I got sick of maintaining a mapping in my head or even in the code to try to determine which fields map to which properties, plus I think this keeps the code generic for each show then the latter parsing passes can take care of the specifics. I imagine if/when we start looking at building out a more robust backend we may revisit this description and perhaps even create separate scrapers for each show.

PR #121, opened 2024-07-06
Feature/move empty row filtering

jjm3x3:
Since actually building contestant data will be more difficult for BB and require more of the "partialContestants" to be exposed we are going to remove that from the wikiFetch and instead into the wikiQuery.

### Screenshots
N/A

## Confirm
- [X] This PR is correctly linked to the relevant issue or milestone.
- [X] This PR has been locally QA'd.

/assign me


CAnthonySiever (approved):
This certainly works, and I'll try to take another look at it tomorrow. `isPartialContestantData` _feels_ a bit clumsy, but that might be due to the nature of working with Wikipedia data for contestant data across different shows without managing our own DB 😅 So this might be as good as it gets. 

I can go ahead and approve, since any improvements here would likely be expanding upon this 👍

jjm3x3:
> This certainly works, and I'll try to take another look at it tomorrow. `isPartialContestantData` _feels_ a bit clumsy, but that might be due to the nature of working with Wikipedia data for contestant data across different shows without managing our own DB 😅 So this might be as good as it gets.
> 
> I can go ahead and approve, since any improvements here would likely be expanding upon this 👍

First off, thanks for taking a look at it and vetting the logic for me so quickly. I totally agree that `isPartialContestantData` _feels_ a bit clumsy. My first response to that would be to ask "is it worse than what we had with `filterEmptyContestants` which technically wasn't really doing that? And to the second part
> that might be due to the nature of working with Wikipedia data for contestant data across different shows without managing our own DB

Yeah, that is kind of how it feels. I am certainly open to better ways of doing this! Especially since, as you will see, I started to make even more changes that make the logic in this function even worse to read 😬

app/utils/wikiQuery.tsx
```diff
@@ -45,15 +53,26 @@ export function getTeamList(contestantData :ITableRowData[]): any {
...
 
     return { props: { runners: contestants }}
 }
+
+export function isPartialContestantData(contestantRowData: ITableRowData): boolean {
+    // Admittedly this is probably a moving target and may come with the baggage
+    // of a doubleNegative since we are often using this as a filter, The idea
+    // is this should return true if we are lacking the necessary/sufficent data
+    // to create a showcontestnat.
+    return contestantRowData.name == null || contestantRowData.name === ""
```
jjm3x3, line 76: Also just checking what folks think about this. Is this code "right" or should it really be
```suggestion
    return !contestantRowData.name
```

jjm3x3: Or, after playing around w/ this for a long time I also realized that some folks prefer something like:
```suggestion
    return Boolean(!contestantRowData.name)
```
to be explicit.

jjm3x3: Or even
```suggestion
    return !Boolean(contestantRowData.name)
```

internette: > Also just checking what folks think about this. Is this code "right" or should it really be

The original is checking for two specific falsy occasions while the new idea is any possible results that equal falsy, i.e. if `contestantData.name` had a value of `0` (the integer, not a string 0), that would also come up as false. 

The change being an improvement depends on what you're trying to achieve. Are you only wanting to check for null or empty string values or all cases of falsy values?

jjm3x3: Huh, it seems like this is where understanding TypeScript a bit better _could_ help us here. So just based on the types we have
```
export interface ITableRowData {
    name: string
    name2: string
    col1: string
    col2: string
    col3: string
    col4: string
    col5: string
}
```
Which says the `row.name` can't be anything that isn't a `:string` (this would include `null` actually, just tried this out locally). Where this starts to get a bit more elusive is just because type checker attempts to ensure this is the case type wise, that doesn't change what is possible during runtime... and that field comes from [this line](https://github.com/jjm3x3/AmazingRaceFantasy/blob/main/app/utils/wikiFetch.tsx#L83)
```
        const name = $row.find('th span.fn').text().trim()
```
which if I had to guess absolutely could return null... though now I am wondering if it would opt to return empty string 🤔 🤔 🤔

internette: Also thought about the name of the function `isPartialContestantData`: Seems like we're only checking if the contestant name is missing. If we're only checking for the name, `hasNoContestantName` or `isMissingContestantName` is a better name for the function.

internette (commented):
Added comments

app/utils/wikiQuery.tsx
internette: Re: your comment
My gut tells me your reply is also true because of server-side rendering. 

Also, I shared how JS generally works to help you with this decision. I'm guessing we can rely on the type-checking to ensure we have a string from Wikipedia and we don't need to do anything fancy if it isn't?

jjm3x3: Sure, now I am just wondering what happens if cheerio isn't able to pull text like we expect does it properly return an empty string which will need to be checked latter, but is at least the right type or is it possible for it to return a null or undefined 😬

internette (approved):

Issue #127, opened 2024-07-13
Add comprehensive testing for the current BB parsing functionality


jjm3x3:
Based on a few outside conversations (both outside of github and outside of the internet or IRL) I am opening this issue so that we can hold ourselves (or myself) accountable for taking on Techdebt in the form of not adding the appropirate testing for the code about to be merged by #125. While I recognize this is a dangerous precident. I am confident that we will be able to come back to this as soon as the season starts rolling and by taking this shortcut now we can get the season rolling with one fewer hurdle.

PR #134, opened 2024-07-27
Bugfix/two small buggs

jjm3x3:
So as soon as the first elimination happened (and as a result the first scoring) it became apparent that we have a couple of bugs. In order of discovery and also in order of effect
1. The status space in the BB25 had text without a space between "Eviction" and "Day" like "EvictionDay" but now there is a white-space so we needed to update the regex which finds a match.
2. Turns out Kenny's last name was originally wrong and so it needed to get updated meaning that every contestant who already had their ranking uploaded had his last name misspelled meaning their contestantRoundListGeneration would break.

### Screenshots
N/A

## Confirm
- [ ] This PR has unit tests scenarios. (arguably this should have at least 1 new test, but I will be leaving that to #127)

/assign me


CAnthonySiever (approved):
👍

PR #128, opened 2024-07-13
126 create contestants page

jjm3x3:
_Notice: this is a WIP since it currently depends on #125_

### Summary/Acceptance Criteria
*Description of what this PR accomplishes*

### Screenshots
*If applicable*

## Confirm
- [ ] This PR has unit tests scenarios. (as in #101 ) we  won't be adding tests here 😅 
- [x] This PR is correctly linked to the relevant issue or milestone.
- [x] This PR has been locally QA'd.

## TODOs
- [x] Merge #125 
- [x] Remove other pages
- [x] Change URLs to be for season 26
  - [x] Remove the `GOOGLE_SHEET_URL` (this will come back for #123)

/assign me

PR #135, opened 2024-07-28
127 add comprehensive testing for the current bb parsing functionality

jjm3x3:
Here are a bunch of unit tests to capture the current behavior of the `getCompetingEntities` function that we are using to parse Big Brother show contestants.

_Note: though technically a bit out of scope, I also updated the ctor for the `Team` type to make a bit more sense and throw a slightly better error_

_Note2: because of a file move the diff for just the tests could be easier to read [here](https://github.com/jjm3x3/AmazingRaceFantasy/compare/356cf802f5fd4edfcb63db3682eb292eddacc101...127-add-comprehensive-testing-for-the-current-bb-parsing-functionality)_

### Screenshots
N/A We are just doing tests here

## Confirm
- [x] This PR has unit tests scenarios. (Now there are 😅)

/assign me


jjm3x3:
I am opening this up even though I know that when the tests run there is a bunch of noisy debug statements because of how the code currently easy. I am opting to leave those in in-case they help indicate a failure also there is a warning in the test output and that is because of a bug that was introduced with #128 I even wrote [this comment](https://github.com/jjm3x3/AmazingRaceFantasy/pull/128/files#r1682050025) calling it out 🫠 Alas here we are, so I have filed #136 in order to address that in the coming weeks, which should fix both the larger bug described in that ticket as well as this warning!

PR #141, opened 2024-08-19
137 add a league standing page

internette:
Tables now show with ranking and sorted

### Screenshots
image

## Confirm

/assign me


app/archive/amazing-race-35/league-standing/page.tsx
```diff
@@ -1,18 +1,20 @@
...
+import { generateContestantRoundScores } from '../../../generators/contestantRoundScoreGenerator'
+import LeagueStandingTable from '../../../components/leagueStandingTable/leagueStandingTable';
 
 export default async function LeagueStanding() {
 
     const dataFetcher = getWikipediaContestantDataFetcher(WIKI_API_URL, "Cast")
-    const contestantScoresComponent = await generateContestantRoundScoreComponent(dataFetcher, CONTESTANT_LEAGUE_DATA)
-
+    const contestantScoresData = await generateContestantRoundScores(dataFetcher, CONTESTANT_LEAGUE_DATA);
+    const contestantScores = contestantScoresData.rounds[contestantScoresData.rounds.length-1].contestantRoundData;
```
jjm3x3, line 11: fft: Is there not a better way of doing this in javascript in 2024?
Like what about with `.slice()`
```suggestion
    const contestantScores = contestantScoresData.rounds.slice(-1)[0].contestantRoundData;
```
or maybe even `.at`
```suggestion
    const contestantScores = contestantScoresData.rounds.at(-1).contestantRoundData;
```
Not exactly sure if either of those have drawbacks, besides just being fresh syntax.

jjm3x3: Also just putting this out there, but this might be a good time to start talking about back-end vs front-end and how to do better templating. I was trying to put as much of this logic in that generator (even including the components) so that the pages are as simple between leagues making the copy and paste job as minimal as possible. Obviously part of that problem is copy and paste, so that obviously has it's own solution. But the other part is if we only want to display the most recent results perhaps it is worth considering if the generator could take that as an argument or an option. WDYT?

app/active/big-brother-26/league-standing/page.tsx
```diff
@@ -0,0 +1,24 @@
+import { WIKI_API_URL } from '../../../leagueConfiguration/BigBrother_26'
+import { CONTESTANT_LEAGUE_DATA } from '../../../leagueData/BigBrother_26'
+import { getWikipediaContestantDataFetcher } from '../../../utils/wikiFetch'
+import { getCompetingEntityList } from "../../../utils/wikiQuery"
+import generateListOfContestantRoundLists from '../../../generators/contestantRoundListGenerator'
+import LeagueStandingTable from '../../../components/leagueStandingTable/leagueStandingTable';
+
+export default async function LeagueStanding() {
+    const dataFetcher = getWikipediaContestantDataFetcher(WIKI_API_URL, "HouseGuests")
+    const contestantScoresData = await generateListOfContestantRoundLists(dataFetcher, CONTESTANT_LEAGUE_DATA, getCompetingEntityList)
```
jjm3x3, line 10: How come we don't use `generateContestantRoundScores`?

jjm3x3 (commented):
Just working through this bit by bit.

app/active/big-brother-26/league-standing/page.tsx
internette: because of the requirement of `team`.  I figured I'll refactor in the follow up ticket when I refactor all of that crazy logic

app/archive/amazing-race-35/league-standing/page.tsx
internette: I agree. There's a bit of overall cleanup we should do. I think after user accounts, we should take the time to refactor and clean up

jjm3x3: Ok, but what about my first suggestions?

app/active/big-brother-26/league-standing/page.tsx
internette: Resolved in a cleanup. Marking as resolved due to the outdated nature of the code

app/components/leagueStandingTable/leagueStandingTable.tsx
```diff
@@ -0,0 +1,37 @@
...
+    type rowState = any[];
+    const tableColNames: rowState = ["Rank", "Name", "Score"];
+    const tableRows: rowState = [];
+    const tableData = {
+        colsNames: tableColNames,
+        rows: tableRows
+    }
+
+    contestantsScores.forEach((contestantData:any) =>{
+        const { content: { props }} = contestantData;
```
jjm3x3, line 18: I think you will find that if you use `generateContestantRoundScores` you wouldn't need to do this

jjm3x3 (commented):
Basically done w/ the review, LMK if you wanna talk about any of these things in more detail.

app/components/leagueStandingTable/leagueStandingTable.tsx
jjm3x3: Sorry, I couldn't stop thinking about this and I wanted to expand on this a bit. Looking a tad deeper, I was realizing what has been done in here to date and technically there are two "functions" called `generateContestantRoundScores` so that would certainly make anyone confused at this suggestion. Moreover they kind of do different things 😅 

_(as a slight aside, I would recommend anyone who finds this message to go take a look and try to tease apart what the difference between the two functions is. It's a bit subtler than you might think, and could serve as an interesting understanding of how code evolves over time)_

_Anyway_... I would probably suggest, depending on our need here, that we use one or both of them and perhaps refactor one or both to be a bit less confusing for the future.

Also ICYMI either one of the functions should result in something that looks a bit like this example
```
League Stnading : {
    "rounds": [
        {
            "round": 5,
            "contestantRoundData": [
                {
                    "name": "Antoinette",
                    "roundScore": 70,
                    "totalScore": 670
                },
                {
                    "name": "Duncan",
                    "roundScore": 80,
                    "totalScore": 660
                }
            ]
        },
        {
            "round": 0,
            "contestantRoundData": [
                {
                    "name": "Duncan",
                    "roundScore": 140,
                    "totalScore": 140
                },
                {
                    "name": "Antoinette",
                    "roundScore": 140,
                    "totalScore": 140
                }
            ]
        }
    ]
}
```

jjm3x3:
I have one more UI/UX question. The screenshot you shared in the description has the contestant `Andrew` and `Jacob` having the same score`, I assume it put `Andrew` first due to "stable" sorting. Shouldn't the positions listed as their rank reflect that fact that they tied though? `Jacob` didn't come in second here, just because their name is second alphabetically.

app/components/leagueStandingTable/leagueStandingTable.tsx
```diff
@@ -0,0 +1,37 @@
...
+        const contestantRoundScores = props.contestantRoundScores;
+        const contestantRoundTotalScore = contestantRoundScores.at(-1).contestantRoundData[0].totalScore;
```
jjm3x3, line 20: This blind indexing, (the `-1` part at least) seems a bit error prone to me. We have a round number, why not use it? Also I think if we start leveraging the `LeagueStanding` and or one of the `generateContestantRoundScores` as mentioned above we won't need the second level of indexing (the `[0]` part).

app/components/leagueStandingTable/leagueStandingTable.tsx
```diff
@@ -0,0 +1,37 @@
...
+        const tableContestantData = [props.contestantName, contestantRoundTotalScore]
```
jjm3x3, line 21: I would recommend this kind of restructuring after the sort especially since we are going to mix datatypes in this array now.

app/components/leagueStandingTable/leagueStandingTable.tsx
```diff
@@ -0,0 +1,37 @@
...
+    tableData.rows.sort((a, b) => {
+        const aScore = a.filter((param:number|string) => typeof param === "number")[0];
+        const bScore = b.filter((param:number|string)=> typeof param === "number")[0];
+        const scoreToReturn = aScore > bScore ? -1 : 1;
+        return scoreToReturn;
+    });
+
+    tableData.rows.map((tableRow, index) => {
+        return tableRow.unshift(index+1);
+    })
```
jjm3x3, line 34: I don't suppose I can convince you that this is a concern of the display layer and so it should go in the component itself 😅

jjm3x3 (commented):
Sorry for diving into this deeper just wanted to do a thorough review. 😅

app/components/baseComponents/components/table/table.module.scss
```diff
@@ -0,0 +1,23 @@
+@use '../../styles/colors' as colors;
```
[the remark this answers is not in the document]
internette: combined them into `:not(.tableHeaderRow:last-child)`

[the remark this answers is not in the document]
internette: Done

app/components/baseComponents/components/table/table.module.scss
```diff
@@ -0,0 +1,23 @@
...
+.tableCell {
+    padding: 0.5rem;
+    text-align: left;
+    &:not(:last-child):not(.tableHeaderRow){
```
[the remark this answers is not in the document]
internette: Combined them into :not(.tableHeaderRow:last-child). We need to ensure we are focusing these styles on all cells that aren't header cells.

app/components/baseComponents/components/table/table.module.scss
```diff
@@ -0,0 +1,23 @@
...
+        border-right: 1px solid colors.$borderGray;
+    }
+}
+
+.tableRow {
+    &:not(:last-child){
+        border-bottom: 1px solid colors.$borderGray;
+    }
+}
```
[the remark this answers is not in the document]
internette: It is and this is a good clean up. I wasn't sure if we would need other styles in `.tableRow` when I was building.

app/components/baseComponents/models/iTable.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+}
+
+export interface iTableFooter {
+    rows: iTableFooterItem[]
+}
+
+export interface iTable {
+    caption?: string, 
+    colsNames: Array<String>,
```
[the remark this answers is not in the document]
internette: We can make it singular. That's a good nit. As for abbreviations, I think it's a habit I've picked up from CSS. I'll spell it out if we want to spell out types.

app/components/baseComponents/components/table/table.tsx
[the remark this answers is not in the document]
internette: Awesome I had never heard of this and it's a good point. Tables regularly have a header and footer, so I added that, but you're totally right to bring YAGNI up. Thanks!

app/components/baseComponents/models/iTable.tsx
```diff
@@ -0,0 +1,16 @@
+export interface iTableFooterItem {
```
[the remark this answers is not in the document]
internette: So, preference is `TableFooterItem` over `tableFooterItem` and `iTabledFooterItem`?

app/components/leagueStandingTable/leagueStandingTable.tsx
```diff
@@ -0,0 +1,37 @@
+import { Table } from "../baseComponents";
+import { getWikipediaContestantDataFetcher } from "@/app/utils/wikiFetch";
+import generateListOfContestantRoundLists from '@/app/generators/contestantRoundListGenerator'
+import { getCompetingEntityList } from "@/app/utils/wikiQuery"
+
+import styles from "./leagueStandingTable.module.scss";
+
+export default async function LeagueStandingTable({ contestantsScores }:{contestantsScores: any}){
+    type rowState = any[];
```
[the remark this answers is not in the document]
internette: Agreed and done

app/components/baseComponents/components/table/table.module.scss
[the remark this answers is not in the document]
jjm3x3: But when I play with this locally it doesn't seem to make a difference either way. And despite not being able to see any visual difference, I also am not sure I understand what this is supposed to do any why it should exclude the header cells.

app/components/baseComponents/models/iTable.tsx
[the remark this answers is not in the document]
jjm3x3: Correct. In other languages I have used it is a pretty strong convention that type names use PascalCaseing (they start with a capital letter). This helps differentiate a type from some other variable. AFIK JavaScript is largely the same at least when it comes to class names, that seems a tad less clear when you introduce TypeScript, but I suspect it still holds. And certainly benefit by maintaining that convention IMHO. The reason for the "i" in the name was a holdover of a thing I brought from C# but I don't think that is nearly as useful here.

app/components/baseComponents/components/table/table.tsx
```diff
@@ -0,0 +1,33 @@
...
+        const headerColumnNames = tableData.columnNames.map(columnValue => <th className={styles.tableCell} scope="col"><strong>{columnValue}</strong></th>)
+        headerRow = <thead><tr className={styles.tableHeaderRow}>{headerColumnNames}</tr></thead>;
+    }
+    const getTableRow = (tableRow:any={})=> {
+        return <tr className={styles.tableRow}>
+            <th className={styles.tableCell}>{tableRow.rank}</th>
+            <td className={styles.tableCell}>{tableRow.name}</td>
+            <td className={styles.tableCell}>{tableRow.totalScore}</td>
+        </tr>
+    }
```
jjm3x3, line 15: Two small comments on this,
- Is this normal to just inline define a function like this? Does this need to be a function, not saying it shouldn't, but if assuming it should, why not create a function outside the actual component definition?
- It seems like the "assumption" that the first argument to this function will have a `rank`, a `name` and a `totalScore`, doesn't that mean that this entire component _(or at least this little)_ inner function is pretty specific to our current use of it and not to the more general idea that the name of the class implies, `Table`?

_Also, not blocking just questions that I think are worth asking._

app/components/leagueStandingTable/leagueStandingTable.tsx
```diff
@@ -0,0 +1,26 @@
+import { Table } from "../baseComponents";
+
+import styles from "./leagueStandingTable.module.scss";
+
+export default async function LeagueStandingTable({ contestantsScores }:{contestantsScores: any}){
+    const tableColumnNames: any[] = ["Rank", "Name", "Score"];
+    const tableData = {
+        columnNames: tableColumnNames,
+        rows: contestantsScores.at(-1).contestantRoundData
```
jjm3x3, line 9: fft: I get that `contestantScores` is kind of a weird catch all data object at the moment... but this blind indexing is a small to me. I wonder if we just get this passed in... and/or add a comment wherever this operation actually takes place. My preference would probably be both, but at least a comment might be nice 😁

app/components/leagueStandingTable/leagueStandingTable.tsx
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    tableData.rows.sort((a: any, b: any) => {
+        const aScore = a.totalScore;
+        const bScore = b.totalScore;
+        const scoreToReturn = aScore > bScore ? -1 : 1;
+        return scoreToReturn;
```
jjm3x3, line 16: nit: I don't think this is technically the `scoreToReturn` it is more like the `sortIndicator`
```suggestion
        const sortIndicator = aScore > bScore ? -1 : 1;
        return sortIndicator;
```

jjm3x3 (changes requested):
This is looking really good, just a bunch more things FYC, but at the end of the day the main thing is we need to make sure the scoring page still works, which it looks like it doesn't because it was broken w/ this change set 😬

app/archive/amazing-race-36/league-standing/page.tsx
```diff
@@ -1,18 +1,20 @@
...
+import { getWikipediaContestantDataFetcher } from '@/app/utils/wikiFetch';
+import LeagueStandingTable from '../../../components/leagueStandingTable/leagueStandingTable';
+import { getTeamList } from '@/app/utils/wikiQuery';
+import { generateContestantRoundScores } from '@/app/generators/contestantRoundScoreGenerator';
 
+export default async function LeagueStandingComponent() {
+    const dataFetcher = getWikipediaContestantDataFetcher(WIKI_API_URL, "Cast");
+    const wikiData = await dataFetcher();
+    const pageData = getTeamList(wikiData);
+    const contestantsScores = await generateContestantRoundScores(pageData, CONTESTANT_LEAGUE_DATA);
```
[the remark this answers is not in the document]
internette: Two things:

1. Let me look into current patterns and best practices because I want to see what others say. I don't want to justify with "Because this is the way front end works" without fact checking and concrete proof.
2. We do this as we move towards reuseable components. With templatization, we should be passing in data into dataless components. This sets us up for easy migration to template pages.

app/generators/contestantRoundScoreGenerator.tsx
```diff
@@ -1,27 +1,22 @@
...
 }
 
-async function generateContestantRoundScores(dataFetcher: () => Promise<ITableRowData[]>, listOfContestantLeagueData: any[]) {
-
-    const wikiContestants = await dataFetcher()
-    const pageData = getTeamList(wikiContestants)
-
+export async function generateContestantRoundScores(dataFetcher: () => Promise<ITableRowData[]>, getCompetitorList: any, listOfContestantLeagueData: any[]) {
+    const wikiData = await dataFetcher();
+    const pageData = getCompetitorList(wikiData);
```
jjm3x3, line 13: fft: I don't love this, mostly because I think there are situations where the "competitor" is actually a team... I would also be inclined to consider what it is doing more abstractly... I don't really have a specific suggestion at this time though 🤷 

lastly  what gets assigned back to the thing currently called `pageData` we should probably come up with a better name for that 🤔

app/generators/contestantRoundScoreGenerator.tsx
```diff
@@ -1,27 +1,22 @@
...
 interface Dictionary<T> {
     [Key: string]: T;
 }
 
-async function generateContestantRoundScores(dataFetcher: () => Promise<ITableRowData[]>, listOfContestantLeagueData: any[]) {
-
-    const wikiContestants = await dataFetcher()
-    const pageData = getTeamList(wikiContestants)
-
+export async function generateContestantRoundScores(dataFetcher: () => Promise<ITableRowData[]>, getCompetitorList: any, listOfContestantLeagueData: any[]) {
```
jjm3x3, line 11: Please add a better type for `getCompetitorList`

jjm3x3 (approved):
Overall I think this is really good. Been a long time coming, but we got there 🙌 ❤️

app/generators/contestantRoundScoreGenerator.tsx
internette: Agreed. I'll add notes in the code to think of a better name

app/generators/contestantRoundScoreGenerator.tsx
internette: Added type :)

app/components/baseComponents/components/table/table.tsx
internette: 1. Yeah if the function is small enough, then yeah it's normal.
2. This is true, but I think for our case, that's the most likely use of tables. We can always iterate on this if we need to.

Issue #114, opened 2024-06-10
Move the header onto each subpage


jjm3x3:
Going along with my designing spirit I was thinking it made sense to put the existing header onto each page. As a result I created #93 in the hopes to achieve that. Unfortuately after #86 was completed by #94 being merged that made it a bit more difficult. And this is because #94 introduced a new `getPages` function in order to populate the `Navigation`. This was only a problem because calling this function from the layout did not work exactly the way that we wanted it to. In fact simply by putting that call in the layout, this actually broke the homepage as well.

jjm3x3:
As much as I would like this to be done ASAP, I think it falls a bit short of being a requirement of the [BB update]() so I am moving it to the part2 milestone.

PR #142, opened 2024-08-20
Added navigation to every page

internette:
Navigator bar will appear on every page now, to make navigating through the site easier

Here's why this works: https://nextjs.org/docs/app/building-your-application/routing/layouts-and-templates

### Screenshots
Scoring Page i.e.
image

## Confirm
- [ ] This PR is correctly linked to the relevant issue or milestone.
- [ ] This PR has been locally QA'd.

/assign me


jjm3x3:
As we talked about in person this doesn't work because of what is detailed in #114 and here is what the looks like
image
_Notice: the archived leagues are missing_

jjm3x3 (approved):
Pretty sure this is just great "as-is" the only things I would like is 
1) To know it is that this works despite #114 proved to be a bit tricky _(for me at least)_
2) Can get a test to assert that we will still get all the archive pages to render. Seems to me like that would be a good test here (one of our checklist items)
3) Along with point 1, I think this solves #114, right I would says we should link them.

PR #164, opened 2025-02-23
Cleanup and format

internette:
Updated eslint to ignore files and ran format

**Out of Scope:**
- Fixing errors around "any" typing. _This will be a follow up PR_

## Confirm
- [x] This PR has unit tests scenarios. (this is n/a)
- [x] This PR is correctly linked to the relevant issue or milestone. (this is n/a)

/assign me


__tests__/utils/teamListUtils.js
```diff
@@ -1,78 +1,78 @@
...
         // Note: being at the end of the array assums that that team should he
         //   eliminated first
         // Note2: rounds are 0 indexed so the first round is 0
 
         // Arrange
         const aTeam = {
-            isInPlay: (round) => true
-        }
-        const teamList = [{}, aTeam]
+            isInPlay: () => true
```
jjm3x3, line 25: I feel like this param should be left in, no?

__tests__/utils/teamListUtils.js
```diff
@@ -1,78 +1,78 @@
...
 
-    it('should be false when the target team is at the end of array and its the second round round', () => {
+    it("should be false when the target team is at the end of array and its the second round round", () => {
 
         // Arrange
         const aTeam = {
-            isInPlay: (round) => true
-        }
-        const teamList = [{}, aTeam]
+            isInPlay: () => true
```
jjm3x3, line 40: Same here

__tests__/utils/teamListUtils.js
```diff
@@ -1,78 +1,78 @@
...
+    });
 
-    it('should be true when the target team is in the 2nd position or second from the end of array and its the first round', () => {
+    it("should be true when the target team is in the 2nd position or second from the end of array and its the first round", () => {
         // Arrange
         const aTeam = {
-            isInPlay: (round) => true
-        }
-        const teamList = [aTeam, {}]
+            isInPlay: () => true
```
jjm3x3, line 54: and here

app/components/baseComponents/components/table/table.tsx
```diff
@@ -1,28 +1,30 @@
...
 
-export default function Table({tableData, tableClassName}:{ tableData: TableData, tableClassName?:String }){
+
+const getTableRow = (tableRow:any={})=> {
+    return <tr className={styles.tableRow}>
+        <th className={styles.tableCell}>{tableRow.rank}</th>
+        <td className={styles.tableCell}>{tableRow.name}</td>
+        <td className={styles.tableCell}>{tableRow.totalScore}</td>
+    </tr>;
+};
```
jjm3x3, line 11: Shouldn't this work go in it's own PR? Or not... just took me a while to understand this diff because I wasn't expecting this here! 🙃

app/components/baseComponents/models/tableData.tsx
```diff
@@ -10,7 +10,7 @@ export interface TableFooter {
 
 export interface TableData {
     caption?: string, 
-    columnNames: Array<String>,
+    columnNames: Array<string>,
```
jjm3x3, line 13: 💟 TIL: [this is why](https://stackoverflow.com/questions/14727044/what-is-the-difference-between-types-string-and-string)

jjm3x3 (commented):
About half way through mostly, looking good so far!

__tests__/utils/teamListUtils.js
internette: Ok, I'll add this rule

app/components/baseComponents/models/tableData.tsx
internette: Ooooo

app/components/baseComponents/components/table/table.tsx
internette: Ahhh sure it could be. But this was under "Cleanup" since you wanted me to move this function from outside the Table function below.

eslint.config.mjs
```diff
@@ -3,37 +3,39 @@ import pluginJs from "@eslint/js";
...
+                ...globals.node,
+                browser: true,
+                module: true
+            }
+        },
+        rules: {
+            "react/react-in-jsx-scope": "off",
+            "react/no-unescaped-entities": "off",
+            "no-unused-vars": "off",
+            "@typescript-eslint/no-unused-vars": "off",
```
jjm3x3, line 27: I think I would like to recommend we don't turn this off, but instead figure out how to use the [argsIgnorePattern](https://eslint.org/docs/latest/rules/no-unused-vars#argsignorepattern) because that way we can be really clear when we don't want it to check, but it will still check by default

eslint.config.mjs
```diff
@@ -3,37 +3,39 @@ import pluginJs from "@eslint/js";
...
-        ],
-        quotes: [
-          "error",
-          "double"
-        ],
-        semi: [
-          "error",
-          "always"
-        ]
-    }
```
jjm3x3, line 37: Wait, I thought you linked these, why are we getting rid of them?

jjm3x3 (commented):
Alright, I went through all the files, but lets talk about these two rules that we are still working out before we merge 😬

eslint.config.mjs
internette: It's deprecated. I plan to implement it correctly in a follow up PR:
https://eslint.org/docs/latest/rules/semi

app/generators/contestantRoundListGenerator.tsx
```diff
@@ -14,35 +13,35 @@ interface Dictionary<T> {
 export default async function generateListOfContestantRoundLists(
     dataFetcher: () => Promise<ITableRowData[]>,
     listOfContestantLeagueData: any[],
-    getCompetingEntityListFunction: (x: ITableRowData[]) => any = getTeamList,
+    getCompetingEntityListFunction: (_x: ITableRowData[]) => any = getTeamList,
```
jjm3x3, line 16: As discussed in person `_` is probably preferable because this isn't even a variable that is "accessible". Of course in order to permit this we will need to update the regex one more time.

jjm3x3 (commented):
So close, just one more lil thing 😅

app/generators/contestantRoundListGenerator.tsx
internette: Done!

jjm3x3 (approved):
🎸

PR #205, opened 2025-03-01
Added unit tests

internette:
Created unit tests for getPages function

## Confirm
- [x] This PR has unit tests scenarios. (This is the unit tests)

/assign me


__tests__/utils/pages.js
```diff
@@ -0,0 +1,117 @@
...
+        return dataObject[filename].leagueConfig;
+      }),
+      checkForSubpages: jest.fn().mockImplementation(filename => {
+        return dataObject[filename].hasOwnProperty('leagueData');
+      })
+    };
+});
+
+describe("pages getPages", () =>  {
+    it("should return appropriate data based on folder structure", () => {
```
jjm3x3, line 94: nit: I think a better name might be something... based on league status, and maybe something about order
```suggestion
    it("should return appropriate data order based on league status", () => {
```

__tests__/utils/pages.js
```diff
@@ -0,0 +1,117 @@
...
+        const pages = pagesModule.getPages();
+        expect(pages.length).toBe(2);
+        expect(pages[0].name).toBe("Current (Big Brother 26)");
+        expect(pages[1].name).toBe("Amazing Race 36");
+    });
+    it("should have subpages based on leagueConfig and leagueData", ()=> {
```
jjm3x3, line 100: nit: could possibly be two seperate functions
- One checks that you have one league per leagueConfig
- Another which checks that you only have `scoring` and `league-standing` pages when there is `leagueData`

app/utils/pages.tsx
```diff
@@ -10,26 +13,36 @@ interface IPage {
     path: string
 }
 
+export function getPathsToMap (){
```
jjm3x3, line 16: nit: shouldn't this have a type
```suggestion
export function getPathsToMap (): string[] {
```

app/utils/pages.tsx
```diff
@@ -10,26 +13,36 @@ interface IPage {
...
+    const pathToLeagueData = path.join(process.cwd(), "app", "leagueConfiguration");
+    const pathsToMap = fs.readdirSync(pathToLeagueData);
+    return pathsToMap;
+}
+
+export function getConfigurationPage (filename:string){
```
jjm3x3, line 22: ... I will explicitly forgive you for not having a type here 😛

app/utils/pages.tsx
```diff
@@ -10,26 +13,36 @@ interface IPage {
...
+    return require(`../leagueConfiguration/${filename}`);
+}
+
+export function checkForSubpages (filename:string){
```
jjm3x3, line 26: nit: let's add a type
```suggestion
export function checkForSubpages (filename:string): boolean {
```

app/utils/pages.tsx
jjm3x3, line 22: Why did you call this function this? Is this ever going to be a page?

jjm3x3 (approved):
Awesome! Love to see us adding tests for things we want to keep alive!

__tests__/utils/pages.js
internette: Done!

app/utils/pages.tsx
internette: Changed name

internette: Hahahaha thank you. This will be typed later lololol

PR #210, opened 2025-03-03
Added leagueConfig to database

internette:
GenerateStaticParams on League Standing now reads from db.

## Confirm
- [x] This PR has unit tests scenarios. (We aren't testing seedDB)

/assign me

### DIFF
--- a/scripts/seedDb.mjs
+++ b/scripts/seedDb.mjs
@@ -4,6 +4,7 @@ import amazingRace37Data from "../app/leagueData/AmazingRace_37.js"
 import bigBrother26Data from "../app/leagueData/BigBrother_26.js"
 import amazingRace35LeagueConfiguration from "../app/leagueConfiguration/AmazingRace_35.js"
 import amazingRace36LeagueConfiguration from "../app/leagueConfiguration/AmazingRace_36.js"
+import amazingRace37LeagueConfiguration from "../app/leagueConfiguration/AmazingRace_37.js"
 import bigBrother26LeagueConfiguration from "../app/leagueConfiguration/BigBrother_26.js"
 import survivor47LeagueConfiguration from "../app/leagueConfiguration/Survivor_47.js"
 import { Redis } from "@upstash/redis"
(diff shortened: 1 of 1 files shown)


scripts/seedDb.mjs
```diff
@@ -51,3 +56,25 @@ async function recreateLeagueData(leagueKeyPrefix, dataRepo) {
         await redis.json.set(leagueKeyPrefix+user.userId, "$", userString)
     }
 }
+
+// Create league configuration data
+
+await recreateLeagueConfigurationData("league_configuration:amazing_race:35:", amazingRace35LeagueConfiguration)
+await recreateLeagueConfigurationData("league_configuration:amazing_race:36:", amazingRace36LeagueConfiguration)
+await recreateLeagueConfigurationData("league_configuration:big_brother:26:", bigBrother26LeagueConfiguration)
+await recreateLeagueConfigurationData("league_configuration:survivor:47:", survivor47LeagueConfiguration)
```
jjm3x3, line 65: nit: I would maybe move these calls up near where we are adding user data.

scripts/seedDb.mjs
```diff
@@ -51,3 +56,25 @@ async function recreateLeagueData(leagueKeyPrefix, dataRepo) {
...
+
+    for (const aKey of leagueConfigurationCursor[1]) { //list of keys
+        redis.del(aKey)
+    }
+
+    const leagueConfigString = JSON.stringify(dataRepo)
+
+    console.log(`Setting league configuration data for '${leagueKeyPrefix}'`);
+    
+    await redis.json.set(leagueKeyPrefix, "$", leagueConfigString)
```
jjm3x3, line 79: I would actually say that we don't want the the key to be the prefix here. The prefix is everything before the final value in the key. In fact I think our "prefixes" here probably aren't prefixes at all, they are just keys.

scripts/seedDb.mjs
```diff
@@ -51,3 +56,25 @@ async function recreateLeagueData(leagueKeyPrefix, dataRepo) {
...
+
+async function recreateLeagueConfigurationData(leagueKeyPrefix, dataRepo) {
+
+    const leagueConfigurationCursor = await redis.scan("0", {match: leagueKeyPrefix+"*"})
+
+    for (const aKey of leagueConfigurationCursor[1]) { //list of keys
+        redis.del(aKey)
+    }
+
+    const leagueConfigString = JSON.stringify(dataRepo)
```
jjm3x3, line 75: Also... I get inserting these as is... but I am also wondering if we shouldn't just do a quick stop gap to change them a bit first so that we are inserting them the way we want the to be saved into the DB. 🤔

jjm3x3 (commented):
Overall it looks ok, but I think there are a few things we should consider.

scripts/seedDb.mjs
internette: I'd love to talk this out with you because I want to hear what you recommend

internette: Update: Did some clean up in this PR
https://github.com/jjm3x3/AmazingRaceFantasy/pull/222

scripts/seedDb.mjs
internette: Updated :)

scripts/seedDb.mjs
internette: Updated and done

jjm3x3 (approved):
First make sure you run this in dev... and then I would recommend you run this in prod, right before or right after merging it.

Issue #196, opened 2025-03-01
Update Contestants page be generated from and to read from DB


jjm3x3:
Now that we have started to develop some muscles around our DB usage. We would like to update our contestants page to read from the DB (this will require a update to our data model) as well as updating our `generateStaticParams` function to be based off the existence of those leagueConfigurations.

Acceptance Criteria
- Update db data model to support leagueConfiguration data
- Update page to read from DB during generation
- Update generateStaticParams to use the DB to determine its existing

internette:
Currently, I have two questions:

1. Should seeding the database exist inside the seedDb function?
	- If yes: We should consider renaming seedDb to seedContestantDb and creating a seedLeagueConfigurationDb. Then create a seedDb function that does both of these functions.
	- If no: I'd recommend renaming the seedDb file to seedContestantDb and having a seedDb file that does both seedContestantDb and seedLeagueConfigurationDb.
2. Should I follow the naming convention suffixed by leagueConfiguration? I.e. show name:season:leagueConfiguration?

Issue #195, opened 2025-03-01
Update generateStaticParams for Scoring and League Standing pages


jjm3x3:
Now that part of the data for these pages is coming from the db (assumes #151 is complete) we should update the `generateStaticParams` for both of these template pages to be based off of information we have in the DB. This seems like it could likely require some updates to our data model such that we can get an accurate picture of what leagues we have data for and will require us to start do reading from redis inside that function.

internette:
<img width="122" alt="Image" src="https://github.com/user-attachments/assets/080c770b-dfe8-48f0-9b7e-c1ca567fa859" />

![Image](https://github.com/user-attachments/assets/50ad13a7-4a6b-4e49-90c7-ec10de7c41b5)

PR #227, opened 2025-03-24
196 update contestants page generate static params be generated from and to read from db

internette:
This migrates generateStaticParams from reading files to reading from the DB

## Confirm
- [x] This PR has unit tests scenarios. (The unit tests shouldn't break)
- [ ] This PR has had it's db migrations run. (ran in a previous PR)

/assign me


app/dataSources/dbFetch.tsx
```diff
@@ -43,6 +43,20 @@ export async function getContestantData(keyPrefix: string): Promise<IContestantD
     return userList
 }
 
+export async function getLeagueConfigurationKeys(): Promise<string[]> {
+    const redis = new Redis({
+        url: process.env.KV_REST_API_URL,
+        token: process.env.KV_REST_API_TOKEN
+    });
+
+    const leagueConfigurationKeys: string[] | null = await redis.keys("league_configuration:*");
```
jjm3x3, line 52: How come we are using `redis.keys()` here as opposed to `redis.scan()` 🤔

jjm3x3 (commented):
Are we planning on updating other pages in other PRs/issues 🤔

app/dataSources/dbFetch.tsx
internette: So, we can do redis.scan. I wasn't sure which we should use.

internette:
> Are we planning on updating other pages in other PRs/issues 🤔

Yep. This ticket was for the last criteria for this: https://github.com/jjm3x3/AmazingRaceFantasy/issues/196

You created an issue for the rest of the generateStaticParams: https://github.com/jjm3x3/AmazingRaceFantasy/issues/195

jjm3x3:
> > Are we planning on updating other pages in other PRs/issues 🤔
> 
> Yep. This ticket was for the last criteria for this: #196
> 
> You created an issue for the rest of the generateStaticParams: #195

Good answer 😁

app/dataSources/dbFetch.tsx
jjm3x3: Where did you see that the "recommendation" is scan? 

What do you mean by
> but that does return all the config data
does it?

internette: The recommendation: https://upstash.com/docs/redis/sdks/py/commands/generic/keys
image

It returns all the League Configuration data as an array of the matching league configurations, based on the matching pattern

jjm3x3: I don't think this last part is true 🤔 
> It returns all the League Configuration data as an array of the matching league configurations, based on the matching pattern

internette: You're right. I was wrong because I forgot about the redis.json.get addition in the function before. I'll switch to scan.

internette:
Updated the PR. When the first item in the array is "0", that means there are no more pages to iterate over. This is seen in the "Response" section of the documentation: https://upstash.com/docs/redis/sdks/py/commands/generic/scan

jjm3x3 (approved):

scripts/seedDb.mjs
```diff
@@ -69,10 +69,16 @@ async function recreateLeagueData(leagueKeyPrefix, dataRepo) {
...
+    while(cursorStart !== "0"){
+        leagueConfigurationCursor = await redis.scan(cursorStart, {match: leagueConfigurationKey+"*"});
+        leagueConfigurationKeys = leagueConfigurationKeys.concat(leagueConfigurationCursor[1]);
+        cursorStart = leagueConfigurationCursor[0];
+    }
 
-    const leagueConfigurationCursor = await redis.scan("0", {match: leagueConfigurationKey+"*"})
-
-    for (const aKey of leagueConfigurationCursor[1]) { //list of keys
+    for (const aKey of leagueConfigurationKeys) { //list of keys
```
jjm3x3, line 81: As discussed offline we don't really need to do a scan here because this is a point read/write. Having said that we will need both a way to delete the existing keys as well as a way to "update" keys (or more likely delete them and rewrite them) to update status.

jjm3x3 (commented):
Just adding a comment to capture what we talked about.

jjm3x3:
I see you merged this, did you run this in prod?

PR #237, opened 2025-03-28
Migrated getPages

internette:
Migrated getPages function to read from db instead of file system

## Confirm
- [ ] This PR has had it's db migrations run.

/assign me

### DIFF
--- a/__tests__/utils/pages.js
+++ b/__tests__/utils/pages.js
@@ -73,7 +73,7 @@ const AmazingRaceLeagueData = ()=> {
             handicap: -80
         }
     ];
-    return { CONTESTANT_LEAGUE_DATA };
+    return CONTESTANT_LEAGUE_DATA;
 }
 
 const dataObject = {
@@ -81,28 +81,36 @@ const dataObject = {
     AmazingRace_36: {leagueConfig: AmazingRaceConfigData(), leagueData: AmazingRaceLeagueData()}
 }
 
-jest.mock("../../app/utils/pages", () => {
+jest.mock("../../app/dataSources/dbFetch", ()=> {
     return {
-        ...jest.requireActual("../../app/utils/pages"),
-        getPathsToMap: jest.fn().mockImplementation(()=> { return Object.keys(dataObject)}),
+        ...jest.requireActual("../../app/dataSources/dbFetch"),
+        getContestantData: jest.fn().mockImplementation((keyPrefix) => {
+            const dataObjKeyArr = keyPrefix.replace(":*", "").split(":");
+            const formattedShowName = dataObjKeyArr[0].split("_").map(word => word.charAt(0).toUpperCase() + word.slice(1)).join("");
+            const showSeason = dataObjKeyArr[1];
+            const dataObjKey =`${formattedShowName}_${showSeason}`;
+            return dataObject[dataObjKey].leagueData;
+        }),
+        getLeagueConfigurationKeys: jest.fn().mockImplementation(()=> {
+            const ARLeagueConfigKey = `league_configuration:${dataObject.AmazingRace_36.leagueConfig.leagueStatus}:amazing_race:36`;
+            const BBLeagueConfigKey = `league_configuration:${dataObject.BigBrother_26.leagueConfig.leagueStatus}:big_brother:26`;
+            return [ARLeagueConfigKey, BBLeagueConfigKey];
+        }),
         getLeagueConfigurationData: jest.fn().mockImplementation(filename => {
             return dataObject[filename].leagueConfig;
-        }),
-        checkForSubpages: jest.fn().mockImplementation(filename => {
-            return Object.prototype.hasOwnProperty.call(dataObject[filename], "leagueData");
         })
-    };
-});
+    }
+})
 
 describe("pages getPages", () =>  {
-    it("should return appropriate data order based on league status", () => {
-        const pages = pagesModule.getPages();
+    it("should return appropriate data order based on league status", async () => {
+        const pages = await pagesModule.getPages();
         expect(pages.length).toBe(2);
         expect(pages[0].name).toBe("Current (Big Brother 26)");
         expect(pages[1].name).toBe("Amazing Race 36");
     });
-    it("should have subpages based on leagueConfig", ()=> {
-        const pages = pagesModule.getPages();
+    it("should have subpages based on leagueConfig", async ()=> {
+        const pages = await pagesModule.getPages();
         const bigBrotherSubpages = pages[0].subpages;
         // Check for leagues based on leagueConfig
         expect(bigBrotherSubpages.length).toBe(1);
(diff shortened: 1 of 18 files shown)


app/utils/pages.tsx
```diff
@@ -13,61 +10,78 @@ interface IPage {
...
-        }
-        //   Path object created
-        const pathObj = {
-            name: pageStrings.friendlyName,
-            subpages: subpages
-        }
-        if(leagueStatus === "active"){
+    for(const leagueConfigurationKey of leagueConfigurationKeys){
+        const pageData:PageInformation = constructPageInformation(leagueConfigurationKey);
+        const showContestantData = await getContestantData(pageData.contestantDataKey);
```
jjm3x3, line 75: As we talked about, I think we should probably not do this call (which will be at least a half a dozen queries), instead we should write another lil function which just gets a single cursors worth to check for the existence of contestant data. If you want to then use that new function inside of `getContetantData` that would be a good idea, and if you would like to do all this in another PR, I would do that first and come back to this to work that in.

jjm3x3 (commented):
Mostly looking pretty good, lots of lil comments, and I would really like to see us not use `getContestantData` directly because we really don't need to query for any actual contestant data, we just want to check if it exists.

__tests__/utils/pages.js
```diff
@@ -73,36 +73,44 @@ const AmazingRaceLeagueData = ()=> {
...
         getLeagueConfigurationData: jest.fn().mockImplementation(filename => {
             return dataObject[filename].leagueConfig;
-        }),
-        checkForSubpages: jest.fn().mockImplementation(filename => {
-            return Object.prototype.hasOwnProperty.call(dataObject[filename], "leagueData");
         })
-    };
-});
+    }
+})
```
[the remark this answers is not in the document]
internette: Yep. Fixed.

__tests__/utils/pages.js
```diff
@@ -73,36 +73,44 @@ const AmazingRaceLeagueData = ()=> {
...
+        getContestantData: jest.fn().mockImplementation((keyPrefix) => {
+            const dataObjKeyArr = keyPrefix.replace(":*", "").split(":");
+            const formattedShowName = dataObjKeyArr[0].split("_").map(word => word.charAt(0).toUpperCase() + word.slice(1)).join("");
+            const showSeason = dataObjKeyArr[1];
+            const dataObjKey =`${formattedShowName}_${showSeason}`;
+            return dataObject[dataObjKey].leagueData;
+        }),
+        getLeagueConfigurationKeys: jest.fn().mockImplementation(()=> {
+            const ARLeagueConfigKey = `league_configuration:${dataObject.AmazingRace_36.leagueConfig.leagueStatus}:amazing_race:36`;
+            const BBLeagueConfigKey = `league_configuration:${dataObject.BigBrother_26.leagueConfig.leagueStatus}:big_brother:26`;
```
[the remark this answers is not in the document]
internette: Agreed. Future enhancement

app/utils/pages.tsx
internette: Committed a fix addressing this :)

app/utils/pages.tsx
```diff
@@ -13,61 +10,78 @@ interface IPage {
...
+    friendlyName: string,
+    contestantDataKey: string
 }
 
-export function getLeagueConfigurationData (filename:string){
-    return require(`../leagueConfiguration/${filename}`);
+function constructPageInformation(leagueConfigurationKey:string){
+    const showKey = leagueConfigurationKey.replace("league_configuration:", "");
+    const params = showKey.replaceAll("_", "-").split(":");
+    const showStatus = params.filter(param => param === "active" || param === "archive")[0];
```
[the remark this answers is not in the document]
internette: Addressed this problem by choosing params based on index instead of filtering

app/utils/pages.tsx
```diff
@@ -13,61 +10,78 @@ interface IPage {
...
+    const showNameAndSeason = params.filter(param => param !== "active" && param !== "archive").join("-");
```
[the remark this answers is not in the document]
internette: Same as above

jjm3x3 (commented):
Nearly done...

app/utils/pages.tsx
```diff
@@ -13,61 +10,76 @@ interface IPage {
...
+interface PageInformation {
+    showStatus: string, 
+    showNameAndSeason: string,
+    friendlyName: string,
+    contestantDataKey: string
 }
 
-export function getLeagueConfigurationData (filename:string){
-    return require(`../leagueConfiguration/${filename}`);
+function constructPageInformation(leagueConfigurationKey:string){
```
jjm3x3, line 20: nit: it might be nice to have an example of what we expect `leagueConfigurationKey` to look like here.

app/utils/pages.tsx
```diff
@@ -13,61 +10,76 @@ interface IPage {
...
+            path: `/${pageInformation.showStatus}/${pageInformation.showNameAndSeason}/league-standing`
+        }
+        subpages.push(scoringSubpage, leagueStandingSubpage);
+    }
+    return subpages;
 }
 
-export function checkForSubpages (filename:string){
-    return fs.existsSync(path.join(process.cwd(), "app", "leagueData", filename));
+function generatePathObj(pageData:PageInformation, hasContestantData:boolean){
```
jjm3x3, line 56: nit: this is usually a big code smell to me. Why pass a boolean when what we really mean is we have two different functions?

app/utils/pages.tsx
```diff
@@ -13,61 +10,76 @@ interface IPage {
...
-        //   Path object created
-        const pathObj = {
-            name: pageStrings.friendlyName,
-            subpages: subpages
-        }
-        if(leagueStatus === "active"){
+    for(const leagueConfigurationKey of leagueConfigurationKeys){
+        const pageData: PageInformation = constructPageInformation(leagueConfigurationKey);
+        const showContestantData = await hasContestantData(pageData.contestantDataKey);
+        const pathObj = generatePathObj(pageData, showContestantData);
```
jjm3x3, line 75: nit: let's rename this variable
```suggestion
        const contestantDataExists = await hasContestantData(pageData.contestantDataKey);
        const pathObj = generatePathObj(pageData, contestantDataExists);
```

jjm3x3 (approved):
A few more nit things, wondering what you think of them. Otherwise I do think we can merge.

app/utils/pages.tsx
internette: This is a great idea. Pushed this up!

app/utils/pages.tsx
internette: We talked about this offline and got a refactor we both were happy about

app/utils/pages.tsx
internette: Committed this change since this was outdated

Issue #229, opened 2025-03-24
The shouldBeScore function needs to support rounds with multi elimination


jjm3x3:
With this latest season of Amazing race one pretty significant thing happened that broke a few different assumptions we were making. What happened is we now need to support what happens when two teams/competing Entities are eliminated at the exact same time, effectively meaning that they are both eliminated in the same round.

How this shows up for this issue is when we have any rounds with two or more eliminations (this has never technically happened up to this point) this function needs to be able to support handling that, both for that current round and for any subsequent rounds.

PR #240, opened 2025-03-30
Feature/insert teamdata into league

jjm3x3:
In the interest of moving round details into the league, which in our current estimation needs to be done in order to complete #229 in order to move the needle in that direction this PR does two things
- Update the ctor to take a list of "`Team`"s (this should be the list of definitive team data as sourced by Wikipedia)
- Move `generateNumberOfRounds` into the league and memoize it so that as we are making modifications as long as we use exactly one `League` object.

### Screenshots
N/A (internal refactor)

## Confirm
- [ ] This PR has unit tests scenarios. (No real behavior has changed just moved around)
- [ ] This PR has had it's db migrations run. (N/A no new data here)

/assign me

### DIFF
--- a/__tests__/models/League.js
+++ b/__tests__/models/League.js
@@ -193,6 +193,68 @@ describe("addContestantRoundScores", () => {
     });
 });
 
+describe("getNumberOfRounds", () => {
+    it("should not ever return Number.MAX_VALUE", () => {
+        //Arrange
+        const teamList = [{eliminationOrder: 1}, {eliminationOrder:2}, {eliminationOrder: Number.MAX_VALUE}];
+        const league = new League(teamList);
+
+        // Act
+        const result = league.getNumberOfRounds();
+
+        // Assert
+        expect(result).not.toBe(Number.MAX_VALUE);
+    });
+
+    it("Should return the number of unique eliminationOrders", () => {
+        //Arrange
+        const teamList = [
+            {eliminationOrder: 1},
+            {eliminationOrder: 1},
+            {eliminationOrder: 3},
+        ];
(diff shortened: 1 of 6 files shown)


__tests__/models/League.js
```diff
@@ -193,6 +193,68 @@ describe("addContestantRoundScores", () => {
     });
 });
 
+describe("getNumberOfRounds", () => {
+    it("should not ever return Number.MAX_VALUE", () => {
+        //Arrange
+        const teamList = [{eliminationOrder: 1}, {eliminationOrder:2}, {eliminationOrder: Number.MAX_VALUE}];
```
internette, line 199: Note: Talked offline about the use of MAX_VALUE. Currently, we use both 0 and MAX_VALUE to indicate that a person is not eliminated. We will discuss offline and make a choice of how to move forward with a single standard. Any refactor will be a future PR.

__tests__/models/League.js
```diff
@@ -193,6 +193,68 @@ describe("addContestantRoundScores", () => {
...
+        const league = new League(teamList);
+
+        // Act
+        const result = league.getNumberOfRounds();
+
+        // Assert
+        expect(result).toBe(2);
+    });
+
+    it("Should not countin Number.MAX_VALUE as a unique number", () => {
```
internette, line 225: Discussed MAX_VALUE above already

internette (approved):

__tests__/models/League.js
jjm3x3: Noted and opened issue #242 to cover this... but definitely out of scope for this change and probably our current work entirely.

PR #243, opened 2025-03-31
Feature/create generator functions to use over static ones

jjm3x3:
Here is another step in the direction that will help us address #229. This step sets us up to not have to call getNumberOfRounds more than once per league generated, even including for the scoring page where we used to make a static call we will now make a call on the instance which instead of adding to the leagues list of `IRound`s it will instead just return a new instance of `IRound[]` but still use the same value for numberOfRounds achieving the exact same buildtime characteristic as before this change, while also making sure that `getNumberOfRounds` can basically be treated as private. 

### Changes to look for
- Made `generateContestantRoundScores` an instance method (in order to leverage a memoized `numberOfRouds`)
- Memoized the `numberOfRounds` (meaning we no longer need to precalculate it)
- Remove all passing around and precalculating of `numberOfRounds`

### Screenshots
N/A (this is a _very_ superficial refactor that may or may not be valuable)

## Confirm
- [ ] This PR has unit tests scenarios. _(All the necessary tests already exist IMHO)_
- [ ] This PR is correctly linked to the relevant issue or milestone. _(Not sure if this goes in a milestone... not sure if this is a change that we want at all)_
- [ ] This PR has had it's db migrations run. _(n/a there is no new data with this PR)_

/assign me

### DIFF
--- a/app/generators/contestantRoundListGenerator.tsx
+++ b/app/generators/contestantRoundListGenerator.tsx
@@ -30,7 +30,7 @@ export default async function generateListOfContestantRoundLists(
     const reverseTeamsList = [...pageData].reverse();
 
     const perfectScoreHandicap = 0;
-    const roundScores: IRound[] = League.generateContestantRoundScores(reverseTeamsList, numberOfRounds, "*perfect*", perfectScoreHandicap);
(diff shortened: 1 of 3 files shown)


jjm3x3:
Closing in favor of #268...

Issue #275, opened 2025-04-27
Update Client based on Login Status


internette:
We are still figuring out what being logged in looks like in the UI, but this issue is a placeholder for indicating something will change

jjm3x3:
I think the DoD here looks like displaying both the users "first name" as presented by Google and also their googleUser Id because then we can go into the backend and associate their googleId w/ our internal userId

PR #279, opened 2025-04-27
Google login setup

internette:
This PR sets up Google login and the API.

### Screenshots
image

## Confirm
- [x] This PR has unit tests scenarios. (The unit tests will be added with https://github.com/jjm3x3/AmazingRaceFantasy/issues/275 )
- [x] This PR has had it's db migrations run. (N/A)

/assign me


app/components/navigation/google-login-btn.tsx
jjm3x3, line 1: Should this have the `"use client";` directive? IIUC the answer is probably _technically_ "no", so long as it is always used by other components which do have the directive But also this component can never properly work as a server side component, right? So in that case it might make sense to be explicit.

app/components/navigation/google-login-btn.tsx
```diff
@@ -0,0 +1,55 @@
...
+        setScriptLoaded(true);
+    };
+
+    const googleLoginRef = useRef(null);
+
+    useEffect(()=> {
+        if(scriptLoaded){
+            const google = window.google;
+            google.accounts.id.initialize({
+                client_id: "708154320268-432vdrg2g0h1652frag5vbu8r8qi4ers.apps.googleusercontent.com",
```
jjm3x3, line 21: Is this what needs to be updated to use a `client_id` that is created under the xfactorleaguesite@gmail.com account?

app/components/navigation/google-login-btn.tsx
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    async function handleLogin(response: Response) {
+        const data = await response.json();
+        console.log(data);
+    }
+
+    return (<>
+        <div ref={googleLoginRef} id="google_login_btn"/>
+        <Script async src="https://accounts.google.com/gsi/client" onLoad={handleScriptLoad}/>
```
jjm3x3, line 53: TIL: `async` is a "prop" you can pass to a React script component 😁

jjm3x3, line 53: Also, wondering should this be capitalized or should it be lower case? [These docs](https://react.dev/reference/react-dom/components/script) seem to have it lower case 🤔

app/api/login/route.ts
```diff
@@ -0,0 +1,21 @@
+import { OAuth2Client } from "google-auth-library";
+import { NextRequest, NextResponse } from "next/server";
+const client = new OAuth2Client();
+
+export async function POST(request: NextRequest) {
+    const body = await request.json();
+    const clientId = process.env.GOOGLE_LOGIN_CLIENT_ID;
```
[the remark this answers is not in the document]
jjm3x3: Even w/o having this set, it does seem to work. So then the question becomes, what is this for 🤔

app/api/login/route.ts
```diff
@@ -0,0 +1,21 @@
...
+        audience: clientId
+    });
+    const payload = ticket.getPayload();
+    const userObj = {
+        email: payload?.email,
+        name: {
+            firstName: payload?.given_name,
+            lastName: payload?.family_name
+        }
+    }
```
jjm3x3, line 19: fft: I guess we probably do always have these props, however I am wondering what this looks like in the long term. How does the site stay logged in? And how/when are we re decoding these tokens? Do we always pass back this info? Is this info we should persist? Not sure any of these need to be answered right now, but we should be careful to make sure that our backend and front end stay in sync and we don't end up with some weird mismatched expectations.

app/components/navigation/google-login-btn.tsx
```diff
@@ -0,0 +1,55 @@
...
+export default function GoogleLoginButton(){
+    const [scriptLoaded, setScriptLoaded] = useState(false);
+    const handleScriptLoad = () => {
+        setScriptLoaded(true);
+    };
+
+    const googleLoginRef = useRef(null);
+
+    useEffect(()=> {
+        if(scriptLoaded){
```
jjm3x3, line 18: Dumb question: if we already have `scriptLoaded` as a dep, then how come we should check here to make sure it actually has?

app/components/navigation/navigation-item.tsx
```diff
@@ -12,9 +12,9 @@ export default function NavigationItem({ inputAttr, labelAttr, listAttr, childEl
         <label 
             htmlFor={inputAttr.id}
             id={labelAttr.id}
-            aria-controls={labelAttr.aria.controls}
+            aria-controls={labelAttr?.aria?.controls}
             data-testid={labelAttr.testId}
             className={labelAttr.classes }>{labelAttr.content}</label>
-        {hasSubpages && <ul className={listAttr.classes}  id={listAttr.id} data-testid={listAttr.testId} aria-hidden={isHidden} hidden={isHidden}>{childElements}</ul>}
+        {hasSubpages && <ul className={listAttr?.classes}  id={listAttr?.id} data-testid={listAttr?.testId} aria-hidden={isHidden} hidden={isHidden}>{childElements}</ul>}
```
jjm3x3, line 18: nit: is there a better way to format this line so we don't invoke a line wrap

app/components/navigation/navigation.tsx
```diff
@@ -26,36 +27,40 @@ export default function Navigation({ pages }: {
...
+                    },
+                    testId: `subpage-${keyName}-label`
+                };
+                const subpageListAttr = {
+                    id: `subpage-${keyName}-dropdown`,
+                    testId: `subpage-${keyName}-dropdown`,
+                    classes: styles["nav-subpages-list"]
+                };
+                return (<li key={`nav-toplevellink-${keyName}`}>
+                    <NavigationItem hasSubpages={hasSubpages }  inputAttr={subpageInputAttr } labelAttr={subpageLabelAttr } listAttr={ subpageListAttr} childElements={page.subpages.map((subpage: ISubpage) => {
```
jjm3x3, line 55: nit: can we find a way to not invoke line wrap?

jjm3x3, line 55: nit: lots of extra spaces... 😬 
```suggestion
                    <NavigationItem hasSubpages={hasSubpages}  inputAttr={subpageInputAttr} labelAttr={subpageLabelAttr} listAttr={subpageListAttr} childElements={page.subpages.map((subpage: ISubpage) => {
```

app/components/navigation/navigation.tsx
```diff
@@ -26,36 +27,40 @@ export default function Navigation({ pages }: {
...
-                    controls: `subpage-${keyName}-dropdown`
-                },
-                testId: `subpage-${keyName}-label`
-            };
-            const subpageListAttr = {
-                id: `subpage-${keyName}-dropdown`,
-                testId: `subpage-${keyName}-dropdown`,
-                classes: styles["nav-subpages-list"]
-            };
-            return (<li key={`nav-toplevellink-${keyName}`} className={styles["nav-toplevel-page"]}>
```
jjm3x3, line 51: Wow! Were we just really not using `.nav-toplevel-page` 😵‍💫. TYVM for getting rid of it!

jjm3x3 (commented):
I feel I am somewhere near 75% of the way through this PR. I also feel obligated to tell you there is a strange bug w/ when the google button decides to load 😬 best I can tell the repro happens somewhere when one navigates to one of the pages for a league then the google button doesn't show up on the "re-rendering" that happens 🤔 

navigating to the page "directly" (by url) doesn't case that... but navigating from the homepage does. May not be a blocker for merging this... but we should defiantly fix it 😬

jjm3x3:
Also why this 😮 
image

app/components/navigation/google-login-btn.tsx
internette: Yes, plus in Vercel

app/components/navigation/google-login-btn.tsx
internette: I don't know if it matters. I capitalized it to indicate it's the next script component as opposed to the HTML script tag (developer legibility).

app/components/navigation/navigation-item.tsx
internette: Yeah. I want to add new lines for props to our ESLint rules, btw

app/components/navigation/navigation.tsx
internette: Np! :)

app/components/navigation/navigation.tsx
internette: Yep!

internette: Will clean up :)

app/api/login/route.ts
```diff
@@ -0,0 +1,21 @@
...
+import { NextRequest, NextResponse } from "next/server";
+const client = new OAuth2Client();
+
+export async function POST(request: NextRequest) {
+    const body = await request.json();
+    const clientId = process.env.GOOGLE_LOGIN_CLIENT_ID;
+    const ticket = await client.verifyIdToken({
+        idToken: body.token,
+        audience: clientId
+    });
```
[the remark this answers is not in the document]
internette: Right, I forgot what you called it so I left it as "ticket". You used another word though.

As for why I used ticket, here's the official google example where they call it a ticket: https://developers.google.com/identity/sign-in/web/backend-auth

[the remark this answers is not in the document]
jjm3x3: Yeah, not sure what is meant by `ticket` here I might recommend something a bit more vague, but appropriate like `authResponse` or I think dotnet usually calls this a `context` or `securityContext` not sure about either of those... I just don't know what a ticket is 🥴

app/components/navigation/google-login-btn.tsx
internette: No, because that is on the navigation component. This may change when we use data from the API, though.

app/components/navigation/google-login-btn.tsx
internette: We only want to run this code if the script is loaded. Otherwise, it will error out

app/components/navigation/google-login-btn.tsx
jjm3x3: Not sure what it being a navigation component has to do with it. The fact that it is using react hooks means that it needs the magic `"use client"` directive.

jjm3x3: Also... moving the conversation to #288 exact link [TBD]()

app/components/navigation/google-login-btn.tsx
jjm3x3: As we discussed off-line I do think it matters, but I guess the difference is if we are using the NextJS component or if we are using the standard HTML5 element. Which also as we discussed we are trying to use the NextJS component so uppercase is probably correct then!

app/api/login/route.ts
jjm3x3: _resolving as the conversation was moved [here](https://github.com/jjm3x3/AmazingRaceFantasy/pull/285#discussion_r2072504820)_

app/components/navigation/google-login-btn.tsx
jjm3x3: Oh... I didn't fully read [the docs](https://react.dev/reference/react/useEffect#reference) on this one so that is my bad 🤦

Issue #277, opened 2025-04-27
Setup Google Application Credentials


internette:
We will need to create an "Application" on the xfactor google account to receive a client id and client secret. We create them at the following URL: https://console.cloud.google.com/

1. Click the hamburger button: 

<img width="214" alt="Image" src="https://github.com/user-attachments/assets/4a4369b4-885c-4508-93cd-410d4b17dee4" />

2. Expand "Solutions" and click "All Solutions" in the navigation that opens

<img width="474" alt="Image" src="https://github.com/user-attachments/assets/9c42a6d0-12db-4827-9edc-d57719369daa" />

3. Under the "Management" section in the body, click "Api & Services"

<img width="723" alt="Image" src="https://github.com/user-attachments/assets/d7a6b6b1-2b1a-4b19-a2b4-1623424874b9" />

4. You should be navigated to a dashboard. In the sidebar, click "Credentials"

<img width="285" alt="Image" src="https://github.com/user-attachments/assets/779521b3-62a4-43b7-b838-c26ca8387708" />

5. You should be navigated to a table with a header that says "Credentials. In the "Credentials" header, click "create Credentials" then "oAuth client ID"

<img width="659" alt="Image" src="https://github.com/user-attachments/assets/6ba9a545-200a-4b3c-942f-9f6de7315825" />

6. Through the form, choose  the following options:
- Application type: Web application
- Authorized JavaScript origins: http://localhost AND  http://localhost:3000. We will need to update this to include https://xfactorleague.com when we go live
- Authorized redirect URIs: http://localhost AND  http://localhost:3000. We will need to update this to include https://xfactorleague.com when we go live

jjm3x3:
Do we really need both `localhost` and `localhost:3000`? 🤔

jjm3x3:
I did most of this, I am unsure what exactly to do next or how to test this 😅

jjm3x3:
Confirmed this is done in person.

PR #288, opened 2025-05-04
Login fe

internette:
This is the front-end part of the API implementation

## Confirm
- [x] This PR has had it's db migrations run.

/assign me

### DIFF
--- a/__tests__/components/googleLoginButton.js
+++ b/__tests__/components/googleLoginButton.js
@@ -1,0 +1,44 @@
+import React from "react";
+import { render, waitFor } from "@testing-library/react";
+import GoogleLoginButton from "../../app/components/navigation/google-login-btn";
+
+const originalGoogle = window.google; // Store original object
+const initializeGoogleMock = jest.fn();
+const requestAccessTokenMock = jest.fn().mockResolvedValue({ access_token: "mock_token" });
+const getClientMock = jest.fn().mockReturnValue({
+    requestAccessToken: requestAccessTokenMock,
+});
+const renderButtonMock = jest.fn().mockReturnValue(()=> {
+    return <div data-testid="google-test-btn">This is my google button</div>
+})
+const mockGoogleAccounts = {
+    id: {
+        initialize: initializeGoogleMock,
+        getClient: getClientMock,
+        renderButton: renderButtonMock
+    },
+};
+
+beforeEach(() => {
+    window.google = { accounts: mockGoogleAccounts };
+});
+
+afterEach(() => {
+    window.google = originalGoogle; // Restore original object
+    jest.clearAllMocks();
+});
+
+describe("GoogleLoginButton Component", () => {
+    it("should render a google login button", () => {
+        const setState = jest.fn();
+        jest.spyOn(React, "useState").mockReturnValue([true, setState]);
+        const { getByTestId } = render(<GoogleLoginButton />);
+        expect(initializeGoogleMock).toHaveBeenCalled();
+        waitFor(()=> {
+            expect(requestAccessTokenMock).toHaveBeenCalled();
+            const googleBtn = getByTestId("google-test-btn");
+            expect(googleBtn).toExist();
+            expect(googleBtn.innerText).toEqual("This is my google button")
+        });
+    });
+});
--- a/app/components/navigation/google-login-btn.tsx
+++ b/app/components/navigation/google-login-btn.tsx
@@ -1,3 +1,4 @@
+"use client"
 import Script from "next/script";
 import { useEffect, useState, useRef } from "react";
 
--- a/app/components/navigation/google-login-btn.tsx
+++ b/app/components/navigation/google-login-btn.tsx
@@ -1,0 +1,55 @@
+import Script from "next/script";
+import { useEffect, useState, useRef } from "react";
+
+interface GoogleLogin {
+    credential: string,
+    select_by: string
+}
+
+export default function GoogleLoginButton(){
+    const [scriptLoaded, setScriptLoaded] = useState(false);
+    const handleScriptLoad = () => {
+        setScriptLoaded(true);
+    };
+
+    const googleLoginRef = useRef(null);
+
+    useEffect(()=> {
+        if(scriptLoaded){
+            const google = window.google;
+            google.accounts.id.initialize({
+                client_id: "708154320268-432vdrg2g0h1652frag5vbu8r8qi4ers.apps.googleusercontent.com",
+                callback: handleCredentialResponse,
+                use_fedcm_for_prompt: true
+            });
+            const parent = googleLoginRef.current;
+            if(googleLoginRef && parent){
+                google.accounts.id.renderButton(parent, {
+                    text: "signin",
+                    size: "medium",
+                    logo_alignment: "left",
+                    shape: "rectangular",
+                    theme: "outline",
+                    type: "standard"
+                });
+            }
+        }
+    }, [scriptLoaded]);
+    
+    function handleCredentialResponse(response:GoogleLogin) {
+        fetch("/api/login", {
+            method: "POST",
+            body: JSON.stringify({ token: response.credential }),
+        }).then(handleLogin);
+    }
+
+    async function handleLogin(response: Response) {
+        const data = await response.json();
+        console.log(data);
+    }
+
+    return (<>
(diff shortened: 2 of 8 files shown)


app/components/navigation/google-login-btn.tsx
jjm3x3, line 1: #### Moving this conversion [from here](https://github.com/jjm3x3/AmazingRaceFantasy/pull/279/files#r2062774406)

> > Should this have the `"use client";` directive? IIUC the answer is probably _technically_ "no", so long as it is always used by other components which do have the directive But also this component can never properly work as a server side component, right? So in that case it might make sense to be explicit.
>
> No, because that is on the navigation component. This may change when we use data from the API, though.

_[recreated](https://github.com/jjm3x3/AmazingRaceFantasy/pull/279/files#r2072778049)_

Not sure what it being a navigation component has to do with it. The fact that it is using react hooks means that it needs the magic `"use client"` directive.

app/components/navigation/navigation-item.tsx
```diff
@@ -1,6 +1,6 @@
 import { useState } from "react";
 import INavigationItem from "@/app/models/INavigationItem";
-export default function NavigationItem({ inputAttr, labelAttr, listAttr, childElements, hasSubpages }: INavigationItem) {
+export default function NavigationItem({ inputAttr, labelAttr, listAttr, childElements }: INavigationItem) {
```
jjm3x3, line 3: 😍😍😍

app/components/navigation/google-login-btn.tsx
```diff
@@ -0,0 +1,55 @@
...
+        setScriptLoaded(true);
+    };
+
+    const googleLoginRef = useRef(null);
+
+    useEffect(()=> {
+        if(scriptLoaded){
+            const google = window.google;
+            google.accounts.id.initialize({
+                client_id: "708154320268-432vdrg2g0h1652frag5vbu8r8qi4ers.apps.googleusercontent.com",
```
jjm3x3, line 22: Shouldn't this come from somewhere else and not be checked in 😅

app/components/navigation/navigation.module.scss
```diff
@@ -16,22 +16,32 @@
...
+    position: relative;
+    text-wrap: nowrap;
+    text-overflow: ellipsis;
+    overflow: hidden;
+    padding-right: 1.75rem;
+    width: 100%;
+}
+.nav-dropdown::after {
+    content: '+';
+    display: block;
```
jjm3x3, line 36: noob: what's this for?

app/components/navigation/navigation.module.scss
```diff
@@ -16,22 +16,32 @@
...
+    position: absolute;
+    top: 50%;
+    right: 0.5rem;
+    transform: translateY(-50%);
```
jjm3x3, line 40: noob: also not sure how you came up with these values especially `top` and `transform` but when I tried to read the docs my eyes went cross-eyed 😬 😵‍💫

app/components/navigation/navigation.tsx
```diff
@@ -26,36 +27,39 @@ export default function Navigation({ pages }: {
...
+                    },
+                    testId: `subpage-${keyName}-label`
+                };
+                const subpageListAttr = {
+                    id: `subpage-${keyName}-dropdown`,
+                    testId: `subpage-${keyName}-dropdown`,
+                    classes: styles["nav-subpages-list"]
+                };
+                return (<li key={`nav-toplevellink-${keyName}`}>
+                    <NavigationItem inputAttr={subpageInputAttr } labelAttr={subpageLabelAttr } listAttr={ subpageListAttr} childElements={page.subpages.map((subpage: ISubpage) => {
```
jjm3x3, line 54: nit: we still have some extra whitespace and line  wrapping.
```suggestion
                    <NavigationItem inputAttr={subpageInputAttr} labelAttr={subpageLabelAtt} listAttr={subpageListAttr} childElements={page.subpages.map((subpage: ISubpage) => {
```

package.json
```diff
@@ -19,6 +19,7 @@
     "node": "^20.9.0"
   },
   "dependencies": {
+    "@types/google.accounts": "^0.0.15",
```
jjm3x3, line 22: nit: can't really figure out what we need this for, but it also seems like the simplest package to pull in 🤔

jjm3x3 (commented):
Think we might want to do something when it comes to checking in that clientId

app/components/navigation/google-login-btn.tsx
jjm3x3: We  talked about this out of band and looks like you addressed it.

app/components/navigation/navigation-item.tsx
internette: You're welcome <3

app/components/navigation/google-login-btn.tsx
internette: I thought so as well, but unsure the best place to put it

app/components/navigation/navigation.module.scss
internette: This is the + sign in the navigation drop down

app/components/navigation/navigation.module.scss
internette: Ah, that's for vertical centering. CSS magic <3

app/components/navigation/navigation.module.scss
jjm3x3: Not the + sign. I am asking about `display: block`

app/components/navigation/google-login-btn.tsx
jjm3x3: Do we have documentation that we are working from to set this up?

package.json
internette: Yeah, this adds the necessary types for the FE `window.google.accounts` typing errors

app/components/navigation/google-login-btn.tsx
jjm3x3: We just going off of [this](https://developers.google.com/identity/sign-in/web/sign-in)

package.json
jjm3x3: I am confused, I don't really see any types being used around that. 🤔

__tests__/components/googleLoginButton.js
```diff
@@ -0,0 +1,44 @@
+import React from "react";
+import { render, waitFor } from "@testing-library/react";
+import GoogleLoginButton from "../../app/components/navigation/google-login-btn";
```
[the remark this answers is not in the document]
internette: Agreed!

app/components/navigation/google-login-btn.tsx
internette: This will be addressed in this issue: https://github.com/jjm3x3/AmazingRaceFantasy/issues/291

app/components/navigation/navigation.module.scss
internette: Ahh, the `display:block` is to allow the element to be styled as if it were a div. MDN's definition: "The element generates a block box, generating line breaks both before and after the element when in the normal flow."

Here's more info: https://developer.mozilla.org/en-US/docs/Web/CSS/display

app/components/navigation/google-login-btn.tsx
jjm3x3: I also just created #293 to at least update this value to work with the account created in #277

app/components/navigation/navigation.module.scss
jjm3x3: I guess, what does that mean for us here? Is there a practical reason we need this, or is this more of a "this is the proper way for us to style this so that we don't shoot ourselves in the foot later"?

__tests__/components/googleLoginButton.js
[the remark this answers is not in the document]
jjm3x3: I love that you agreed w/ this and then didn't do anything 😛

jjm3x3 (approved):
Alright, so this is my final review:
I think what we have here is a good enough start IMO. This can block future development and while there are some weird bugs and behaviors (like the one where the button sometimes disappears, which we should log an issue for).

I do think this is good enough for now.

One other thing I think we should at least log an issue for is one that captures the conversation we were having offline about "preview" versions of the site not being able to perform login since their domains need to be allow listed and that is not something that can even be automated, even if we wanted to.

PR #307, opened 2025-07-06
305 store login token clientside

internette:
### Summary/Acceptance Criteria
This creates a session with the session id stored within the NextJS app. Will write encryption and decryption in next PR

## Confirm
- [x] This PR has unit tests scenarios.
- [x] This PR is correctly linked to the relevant issue or milestone.
- [x] This PR has been locally QA'd.
- [x] This PR has had it's db migrations run.

/assign me

PR #310, opened 2025-07-06
Add session encryption

internette:
Follow up to https://github.com/jjm3x3/AmazingRaceFantasy/pull/307 adding encryption ~and decryption~ of sessions

Decryption will be done in the logout issue

## Confirm
- [x] This PR has had it's db migrations run.

/assign me


__tests__/setupTests.js
```diff
@@ -1,3 +1,7 @@
 // react-testing-library renders your components to document.body,
 // this adds jest-dom's custom assertions
 import "@testing-library/jest-dom";
+import { TextDecoder, TextEncoder } from "util";
```
internette, line 4: Added because JOSE package is written in commonjs and it's submodules need to be imported for jest

__tests__/api/login/route.js
```diff
@@ -60,6 +69,7 @@ describe("POST", () => {
         };
         const LoginResponse = await POST(requestMock);
         const body = await LoginResponse.json();
+        console.log(body);
```
internette, line 72:
```suggestion
```

jjm3x3 (commented):
Just something that I noticed, while I get my dev environment working again 😓

next.config.js
```diff
@@ -1,4 +1,6 @@
 /** @type {import('next').NextConfig} */
-const nextConfig = {};
+const nextConfig = {
+    transpilePackages: ["jose"]
+};
```
jjm3x3, line 4: What's this for?

jest.config.mjs
```diff
@@ -9,7 +9,9 @@ const createJestConfig = nextJest({
 const config = {
     // Add more setup options before each test is run
     // setupFilesAfterEnv: ['<rootDir>/jest.setup.js'],
-
+    moduleNameMapper: {
```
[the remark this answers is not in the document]
jjm3x3: Seems like this might be another good reason to switch to Auth.js (or NextAuth.js) maybe we create an issue for it an we can determine when/if we make that conversation.

jjm3x3 (commented):
Would love to see the tests cleaned up and I would like to get this working locally, but the code change looks good 👍

jest.config.mjs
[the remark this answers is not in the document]
internette: Yep. That's what my thought was exactly.

jjm3x3 (commented):
I think it would be good if we updated the readme either copying some of the details from [the next document on auth](https://nextjs.org/docs/app/guides/authentication#1-generating-a-secret-key) _I assume that's how you set this up_, or at least links to it.

This can be a separate PR, if you like.

app/api/session/session.tsx
```diff
@@ -1,6 +1,32 @@
...
+const sessionSecretKey = process.env.SESSION_SECRET;
+const encodedKey = new TextEncoder().encode(sessionSecretKey);
+
+export async function encrypt({ envelope, exp }:{ envelope: JWTPayload, exp: number }) {
+    return new SignJWT(envelope)
+        .setProtectedHeader({ alg: "HS256" })
+        .setIssuedAt()
+        .setExpirationTime(exp)
+        .sign(encodedKey)
+}
```
jjm3x3, line 14: Having worked and played w/ this now, I would ask if we could add some way of error-ing out earlier than when we try to sign. Even right before would give us the control to throw a more sensible error for the next developer.

app/api/session/session.tsx
```diff
@@ -1,6 +1,32 @@
...
+export async function decrypt(session: string | undefined = "") {
+    try {
+        const { payload } = await jwtVerify(session, encodedKey, {
+            algorithms: ["HS256"],
+        })
+        return payload
+    } catch (error) {
+        console.log("Failed to verify session")
+    }
+}
```
jjm3x3, line 25: I am also realizing that while I am sure we want this function... we don't have any current use of it. Might make sense to put in a follow up where we at least test it.

app/api/session/session.tsx
jjm3x3, line 1: I am also a bit curious of the location of this file. Is the idea in having it under the `app/api/` directory somewhere because this is code that is exclusively used by the API?

jjm3x3 (commented):
And one last thing, I think I got this working but since we don't really have decrypt hooked up at all it's hard to test 😅 

That said, I believe this is my final review so I am prepared to own the decisions going forward, but I do think there are some changes we would like to make here and not merge this exactly as is.

app/api/session/session.tsx
internette: Added an error

app/api/session/session.tsx
internette: Removed

app/api/login/route.ts
```diff
@@ -28,11 +28,7 @@ export async function POST(request: NextRequest) {
         }
 
         const response = NextResponse.json(userObj);
-        await createSession({
-            response,
-            exp: payload.exp,
-            session_id: body.token
-        });
+        await createSession({response, envelope: body.envelope, exp: payload.exp});
```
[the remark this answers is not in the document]
internette: I'm sure this will change when we switch to Auth.js, so for now I think this is ok

app/api/session/session.tsx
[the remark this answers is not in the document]
jjm3x3: I agree, plus that would make sure we are not on the hook for possibly "leaking" `googleId`s. The problem with we don't currently have an association between `googleId`s and our userIds, and in the meantime these session tokens aren't really valuable w/o having a sub claim because we need to know who a user is. 😅

README.md
```diff
@@ -14,6 +14,12 @@ We have two key phrases that we use throughout the app:
 1. **Competing Entity**: This is used for any competitor (individual or teams) in a show
 2. **Contestant**: This is used for any competitor in a league
 
+This app requires setting a local environment variable. Here's how to set up this variable: 
+1. Generate a local secret key. One way to do this is to run `openssl rand -base64 32`
```
jjm3x3, line 18: nit: these should all be `1.` so that they never need to be renumbered.

```suggestion
1. Generate a local secret key. One way to do this is to run `openssl rand -base64 32`
1. In your `.env.development.local`, create a variable named `SESSION_SECRET`
1. Set this variable equal to the value you generated in step 1.
1. Stop and start your app.

```

PR #343, opened 2025-08-09
323 log out

jjm3x3:
So far the expectation of this PR is to "draw the rest of the owl" on logout. Once complete users should be able to "logout" of the site effectively meaning they no longer have an active usable session with the site.

### Screenshots
Before logging in:
image

After logging in:
image

_then it goes back to the before login_

## Confirm
- [ ] This PR has had it's db migrations run. (N/A no new data here)

### TODO::
- [x] Determine how to do tests where we can mock fetch
  - [x] Determine how to get through working with `SessionProvider` in a unit test
    - _worked through this by mocking the react hook entirely which should be fine for a unit test_
- [x] Verify the change to navigation-item is correct
- [ ] ~Figure out how to mock next router~ (Moved to #345)
    - [ ] ~this enables us to add a `push` redirect on-logout forcing the client to go to root~ (Moved to #345)
- [x] Verify that we want to update the api response to be 205 (this could change scope on #341 a lil bit)

/assign me

### DIFF
--- a/app/components/navigation/navigation-item.tsx
+++ b/app/components/navigation/navigation-item.tsx
@@ -10,9 +10,7 @@ export default function NavigationItem({
 }: INavigationItem) {
     const [isExpanded, setIsExpanded] = useState(false);
     useEffect(()=> {
-        if(navigationClose){
-            setIsExpanded(false);
-        }
+        setIsExpanded(false);
     }, [navigationClose]);
     return (<>
         <input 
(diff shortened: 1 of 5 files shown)


app/components/navigation/navigation-item.tsx
```diff
@@ -10,9 +10,7 @@ export default function NavigationItem({
 }: INavigationItem) {
     const [isExpanded, setIsExpanded] = useState(false);
     useEffect(()=> {
-        if(navigationClose){
-            setIsExpanded(false);
-        }
+        setIsExpanded(false);
```
jjm3x3, line 13: Talked this change out over dinner, and this made sense to both of us.

internette (approved):

PR #406, opened 2025-12-06
Sometimes Form Submission Fails Fix

internette:
Updated exp to be max age of session, not absolute exp date. 

## Confirm
- [ ] This PR has unit tests scenarios. (N/A: using pre-existing one)

### DIFF
--- a/app/api/login/route.ts
+++ b/app/api/login/route.ts
@@ -37,10 +37,10 @@ export async function POST(request: NextRequest) {
         }
 
         const response = NextResponse.json(userObj);
-        const maxSessionAge = payload.exp - payload.iat;
         await createSession({
             envelope: body.envelope, 
-            exp: maxSessionAge, 
+            exp: payload.exp,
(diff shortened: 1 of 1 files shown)


app/api/login/route.ts
```diff
@@ -37,9 +37,10 @@ export async function POST(request: NextRequest) {
         }
 
         const response = NextResponse.json(userObj);
+        const maxSessionAge = payload.exp - payload.iat;
         await createSession({
             envelope: body.envelope, 
-            exp: payload.exp, 
+            exp: maxSessionAge,
```
jjm3x3: This field/param is used to set both the `exp` claim in the session value, (which should be the same) and also the `maxAge` property on the cookie. So we can't just make this `maxAge` becuase then we will be breaking the `exp` claim on the token itself.

jjm3x3 (approved):
Alright, looks good now! :shipit:

PR #419, opened 2025-12-29
Implement Create Account API

internette:
This essentially copies the login API because the login API is essentially a create account endpoint. The following differences exist:
1. I duplicated the `writeGoogleUserData` with `writeGoogleUserDataWithId`. This is because the login endpoint uses this function. This will change after the login button actually uses the new login endpoint (when that's created).
2. I imported the NodeJS package `crypto` to generate the uuid. This package has a built in cache of 128 unique values. I know we will be moving away from Node when we move the API to it's own repo, so I figured this is good for now. You can learn more about this package here: https://nodejs.org/docs/latest-v14.x/api/crypto.html#crypto_crypto_randomuuid_options

~~Note: I didn't include a check for pre-existing users. We can discuss this if we want to include this check. I thought of checking if the google user ID was a pre-existing entry in our db, but figured we wouldn't want to prevent signup from that alone since that would cause guest friction.~~

Jacob and I discussed offline and agree the above note should be addressed in this PR. As a result, I've updated this PR to include the required changes.

## Confirm

### DIFF
--- a/app/api/account/route.ts
+++ b/app/api/account/route.ts
@@ -36,32 +36,31 @@ export async function POST(request: NextRequest) {
     const existingGoogleUser = await getUser(googleUserId);
     if (existingGoogleUser){
         return NextResponse.json({"error": "User already exists with the provided google user id"}, {status: 409});
-    } else {
-        const uuid = randomUUID();
-        const userDbObj = {
-            googleUserId: googleUserId,
-            userId: uuid
-        }
-        writeGoogleUserDataWithId(userDbObj);
-
-        // Data to send to the front end
-        const userObjForClient = {
-            email: payload?.email,
-            name: {
-                firstName: payload?.given_name,
-                lastName: payload?.family_name
-            },
-            googleUserId: googleUserId,
-            userId: uuid
-        }
-        const response = NextResponse.json(userObjForClient);
-        await createSession({
-            envelope: body.envelope, 
-            exp: payload.exp,
-            iat: payload.iat, 
-            sub: uuid,
-            response
-        });
-        return response;
     }
+    const uuid = randomUUID();
+    const userDbObj = {
+        googleUserId: googleUserId,
+        userId: uuid
+    }
+    writeGoogleUserDataWithId(userDbObj);
+
+    // Data to send to the front end
+    const userObjForClient = {
+        email: payload?.email,
+        name: {
+            firstName: payload?.given_name,
+            lastName: payload?.family_name
+        },
+        googleUserId: googleUserId,
+        userId: uuid
+    }
+    const response = NextResponse.json(userObjForClient);
+    await createSession({
+        envelope: body.envelope, 
+        exp: payload.exp,
+        iat: payload.iat, 
+        sub: uuid,
+        response
+    });
+    return response;
 }


app/api/create-account/route.ts
jjm3x3: FFT: calling this create-account is redundant (and perhaps even "wrong") if we were be trying to build a strictly RESTfull API. In that case we would just call it `/api/account/route.ts` and then we can define the `POST` method which would imply account creation. Not a huge deal, but maybe worth considering before we get too deep into things.

Having said all that, this is one of those benefits of being a "mono-repo" 😁 even if we wanted to just change that name next month we should be able to do so relatively painlessly and in one fell swoop.

app/dataSources/dbFetch.tsx
```diff
@@ -116,6 +116,18 @@ export async function writeGoogleUserData (googleUserId: string){
...
+export async function writeGoogleUserDataWithId (userDbObj: {googleUserId: string, userId: string}){
+    const redisOptions = {
+        url: process.env.KV_REST_API_URL,
+        token: process.env.KV_REST_API_TOKEN
+    };
+    const redis = new Redis(redisOptions);
+
+    // Post to DB
+    const leagueConfigString = JSON.stringify(userDbObj)
+    await redis.json.set(`user:${userDbObj.userId}`, "$", leagueConfigString)
```
jjm3x3: I think we still want to write these in terms of `googleUserId` because of how this collection is being used. So I am temped to block on this 😬 I think at least we should have a conversation before merging this even if we decide to do nothing else.

app/api/create-account/route.ts
```diff
@@ -0,0 +1,61 @@
...
+            writeGoogleUserDataWithId(userDbObj);
+
+            // Data to send to the front end
+            const userObjForClient = {
+                email: payload?.email,
+                name: {
+                    firstName: payload?.given_name,
+                    lastName: payload?.family_name
+                },
+                googleUserId: googleUserId
```
jjm3x3: I almost wonder if we want to just add a temporary claim (like one that we just define right here right now for some amount of time while we transition between these two user ids) to the session token, that way we know this can't just be tampered with. WDYT?

__tests__/api/create-account/route.js
```diff
@@ -0,0 +1,193 @@
...
+            CreateAccountResponse = await POST(requestMock);
+        } catch (testError) {
+        // Assert
+            expect(testError).toBeFalsy(); // should not throw
+        }
+
+        expect(CreateAccountResponse).not.toBeNull();
+    });
+
+    it("should catch an exception when one is thrown during verification and return a 401", async () => {
```
jjm3x3: nit: isn't this test basically the same as the one before?

app/api/account/route.ts
internette: Renamed to account, as we discussed offline

__tests__/api/account/route.js
```diff
@@ -0,0 +1,193 @@
...
+        })
+    }
+});
+
+jest.mock("crypto", ()=> {
+    return {
+        ...jest.requireActual("crypto"),
+        randomUUID: jest.fn().mockImplementation(()=> {
+            return "google-user-uuid-mock"
+        })
```
internette: We need to mock this module so that the test will match the uuid in the response and not randomly generate every time. If we do, it's not easy to test

__tests__/api/account/route.js
```diff
@@ -0,0 +1,193 @@
...
+});
+
+describe("POST", () => {
+    it("should return the mocked access token", async () => {
+        const requestMock = {
+            json: async () => (testRequestPayload),
+        };
+        const CreateAccountResponse = await POST(requestMock);
+        const body = await CreateAccountResponse.json();
+        expect(verifyIdTokenMock).toHaveBeenCalledWith({
```
jjm3x3: nit: This should probably start w/ a lowercase 'c' to maintain consistency
 
```suggestion
        const createAccountResponse = await POST(requestMock);
        const body = await createAccountResponse.json();
        expect(verifyIdTokenMock).toHaveBeenCalledWith({
```

jjm3x3: Literally everywhere 😩

__tests__/api/account/route.js
```diff
@@ -0,0 +1,193 @@
...
+        expect(CreateAccountResponse.status).toBe(401);
+    });
+
+    it("should successfully call Redis", async ()=> {
+        const request = {
+            json: async () => (testRequestPayload),
+        };
+        const CreateAccountResponse = await POST(request);
+        await CreateAccountResponse.json();
+        const expectedRedisResponse = [ "user:123googleTestId", "$", "{\"googleUserId\":\"123googleTestId\",\"userId\":\"google-user-uuid-mock\"}" ];
```
jjm3x3: As discussed this would be a much better test if we didn't include the "fake guid" in the matching here.

__tests__/api/account/route.js
```diff
@@ -0,0 +1,193 @@
...
+        const getAccountExistsPayloadMock = jest.fn().mockImplementation(()=> {
+            return testAuthDataAccountExists
+        });
+        verifyIdTokenMock = jest.fn().mockImplementation(()=> {
+            return {
+                getPayload: getAccountExistsPayloadMock
+            }
+        });
+
+        const { POST } = await import ("../../../app/api/account/route");
```
jjm3x3: What's this about?

app/api/account/route.ts
```diff
@@ -0,0 +1,69 @@
...
+        console.warn(`Issue verifying google identity ${authError}`);
+
+        return NextResponse.json({"error": unauthenticatedErrorMessage}, {status: 401});
+    }
+    
+    const payload:TokenPayload | undefined = authResponse !== null ? authResponse.getPayload() : undefined;
+
+    if(!payload){
+        return NextResponse.json({ "error": badGatewayErrorMessage }, { status: 502 });
+    } else {
```
jjm3x3: nit: now that we have inverted the condition, we can remove this else entirely and flatten our code! 🙌

app/api/account/route.ts
```diff
@@ -0,0 +1,69 @@
...
+import { unauthenticatedErrorMessage, badGatewayErrorMessage, missingBodyErrorMessage, malformedBodyErrorMessage } from "@/app/api/constants/errors";
+
+export async function POST(request: NextRequest) {
+    const client = new OAuth2Client();
+    const body = await request.json();
+    if(!body.token){
+        return NextResponse.json({"error": missingBodyErrorMessage}, {status: 400});
+    } else if (body.token.trim() === "" || (/^[0-9a-zA-Z=.]+/g).test(body.token) === false){
+        return NextResponse.json({"error": malformedBodyErrorMessage}, {status: 400});
+    }
```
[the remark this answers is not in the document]
internette: Done!

app/api/account/route.ts
internette: Done!

__tests__/api/account/route.js
```diff
@@ -0,0 +1,229 @@
...
+    });
+
+    it("should successfully post to Redis", async ()=> {
+        const request = {
+            json: async () => (testRequestPayload),
+        };
+        const CreateAccountResponse = await POST(request);
+        await CreateAccountResponse.json();
+        
+        const expectedRedisResponse = [ "user:123googleTestId", "$", expect.stringContaining("{\"googleUserId\":\"123googleTestId\"")];
```
jjm3x3: Beautiful 😁

PR #444, opened 2026-01-30
Feature/add login page4

### LINKED ISSUE #359: Create new create account page
After the new login page (created with #358) we would like to add a link to it, which will say something like "Create account" and will link to a new Create Account page where there will be a different* Google login in button (perhaps changed text to say "Create Account") which will call the create endpoint (created in #357)

_* it should be a different button long term since it should be stylized differently if we can and it should call a different endpoint, however since both existing endpoints are likely very similar we could always just use the same button for now and move the work to #360_

jjm3x3:
Finally the moment we have all been waiting for! this finally resolves #359 by adding a new page and then linking between the login and the create account page.

### Screenshots

_The login page after adding a link:_
image

_The *new* create account page:_
image

## Confirm
- [ ] This PR has unit tests scenarios. (N/A this is all presentation)
- [ ] This PR has had it's db migrations run. (N/A no new data)
- [ ] Update documentation. (N/A not quite the kind of thing that needs docs)

### DIFF
--- a/app/create-account/page.tsx
+++ b/app/create-account/page.tsx
@@ -9,7 +9,7 @@ export default async function CreateAccount() {
             <div className="text-center">
                 <h3 className="text-xl">Create Account</h3>
                 <GoogleCreateButton classes={styles.google_create_btn} />
-                <p>Already have an account? <Link className="standard-link" href={"/login"}>Login</Link></p>
+                <p>Already have an account? <Link className="standard-link" href={"/login"}>Login</Link>.</p>
             </div>
         </div>
     );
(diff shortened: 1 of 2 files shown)


app/create-account/page.tsx
```diff
@@ -0,0 +1,16 @@
+import Link from "next/link";
+import GoogleCreateButton from "@/app/components/baseComponents/googleButton/createButton";
+import styles from "./styles.module.scss"
+
+export default async function CreateAccount() {
+
+    return (
+        <div className="grid place-items-center h-screen">
+            <div className="text-center">
+                <p className="text-xl">Create Account</p>
```
internette:
```suggestion
                <h3 className="text-xl">Create Account</h3>
```

app/login/page.tsx
```diff
@@ -8,7 +9,7 @@ export default async function Login() {
             <div className="text-center">
                 <p className="text-xl">Login</p>
```
internette:
```suggestion
                <h3 className="text-xl">Login</h3>
```

PR #463, opened 2026-03-02
Toggle elimination status

internette:
Instead of adding a new page, I added an elimination status toggle. We had to move some of the content into a separate component since we are utilizing `useState`, which requires `use client`. I kept the client-side content rendered to a minimum. I also added some basic tests for this page.

### Screenshots
image
image

## Confirm
- [ ] This PR has had it's db migrations run. (n/a)
- [ ] Update documentation. (n/a)

### DIFF
--- a/__tests__/components/baseComponents/checkboxToggle.js
+++ b/__tests__/components/baseComponents/checkboxToggle.js
@@ -17,7 +17,7 @@ describe("CheckboxToggle", ()=> {
         expect(getByTestId("test-checkboxToggle-test").nextElementSibling).toBe(getByTestId("test-checkboxToggle-label-test"));
     });
 
-    it("should toggle the checkbox", ()=> {
+    it("should toggle the checkbox", async ()=> {
         const { getByTestId } = render(<CheckboxToggle labelText="Test Label"
                                                         id="test"
                                                         checkboxPosition="left"


app/[showStatus]/[showNameAndSeason]/contestants/contestantsPageContent.tsx
Comment (jjm3x3) [inline app/[showStatus]/[showNameAndSeason]/contestants/contestantsPageContent.tsx]: nit/fft: I don't love the idea of calling this "PageContent" that doesn't feel like a component... that feels like a page which we just had to "turn into a component" so that we could then add a `"use client"` directive. Maybe let's give this a more component sounding name.

jjm3x3 (commented):
As it is and what it achieves is a good start, but I do think there are some things we should consider.

app/components/baseComponents/components/inputs/checkboxToggle/checkboxToggle.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export default function CheckboxToggle({ 
+    labelText,
+    toggleHandler,
+    checkboxPosition,
+    id
+}:{
+    labelText: string,
+    toggleHandler: () => void,
+    checkboxPosition: "left" | "right",
```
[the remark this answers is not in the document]
internette: Good call out!

app/[showStatus]/[showNameAndSeason]/contestants/teamListWithToggle.tsx
Comment (internette) [reply app/[showStatus]/[showNameAndSeason]/contestants/teamListWithToggle.tsx]: After revamping the code to use TeamList, I renamed this component as "TeamListWithToggle". I kept it in the Contestants page since this is the only page that uses it. We can revisit this later.

app/[showStatus]/[showNameAndSeason]/contestants/page.tsx
```diff
@@ -34,27 +33,20 @@ export default async function Contestants({ params }: {
     const wikiContestants = await getWikipediaContestantData(wikiApiUrl, castPhrase);
 
     const final = parseEntities(wikiContestants, showName);
+    const randomizedContestants = [...final].sort(() => Math.random() - 0.5);
+
+    // This is a workaround to ensure the data is serializable and can be passed to a client-side component.
+    // To understand better, see: https://stackoverflow.com/questions/77091418/warning-only-plain-objects-can-be-passed-to-client-components-from-server-compo
+    const parsedFinal = JSON.parse(JSON.stringify(randomizedContestants));
```
Comment (jjm3x3) [inline app/[showStatus]/[showNameAndSeason]/contestants/page.tsx]: Does leaving this this way make this SSRed?

PR #476, opened 2026-03-20
466 implement elimination status on scoring page

internette:
This PR enables users to toggle from the default shuffled contestants ranking to their own, achieving the goal of seeing one's ranking without spoilers. This PR also includes support for toggling elimination status.

### Screenshots
image
image
image

## Confirm

### DIFF
--- a/app/models/ContestantRoundData.tsx
+++ b/app/models/ContestantRoundData.tsx
@@ -1,4 +1,4 @@
-export type ContestantRoundData = {
+export interface ContestantRoundData {
     name: string
     roundScore: number
     totalScore: number
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
         "react": "^19.2.1",
         "react-dom": "^19.2.1",
         "sass": "^1.77.6",
+        "storybook": "^10.3.3",
         "tailwindcss": "3.3.3",
         "typescript": "5.2.2",
         "zod": "^3.25.71"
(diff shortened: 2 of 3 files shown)


app/[showStatus]/[showNameAndSeason]/contestants/teamListWithToggle.tsx
```diff
@@ -1,31 +1,80 @@
...
+        value: "default",
+        text: "Select a contestant",
+        id: "default",
+        key: "default",
+        teamList: contestantsData
+    }
+    const [selectedContestant, setSelectedContestant] = useState(defaultDataSelectOption);
+    const [selectOptions, setSelectedOptions] = useState([defaultDataSelectOption]);
+    const league = new League(contestantsData);
+    useEffect(() => {
```
Comment (jjm3x3) [inline app/[showStatus]/[showNameAndSeason]/contestants/teamListWithToggle.tsx]: QQ: how come this is in a `useEffect`? That is for when stuff is going to take a while computationally, right? Is that a concern here? If so why don't we do more of this on the sever?

jjm3x3 (commented):
Mostly interested in how are coming up those team lists for league contestants... seems like something we should do on the server instead of on the client.

app/[showStatus]/[showNameAndSeason]/contestants/teamListWithToggle.tsx
Comment (internette) [reply app/[showStatus]/[showNameAndSeason]/contestants/teamListWithToggle.tsx]: My previous comment is no longer true. Offline, Jacob and I agreed that when the logic impacts performance, we will move this logic to an API endpoint

app/[showStatus]/[showNameAndSeason]/contestants/teamListWithToggle.tsx
```diff
@@ -2,30 +2,71 @@
...
+    const playerDataSelectOptions = playerData.map(contestant => {
+        const computerFriendlyName = contestant.name.toLowerCase().replace(/\s/g, "-")
+        return {
+            key: computerFriendlyName,
+            text: contestant.name,
+            value: computerFriendlyName,
+            id: computerFriendlyName,
+            teamList: league.getTeamList(contestant.ranking)
+        }
+    });
```
Comment (jjm3x3) [inline app/[showStatus]/[showNameAndSeason]/contestants/teamListWithToggle.tsx]: nit/fft: I still wouldn't do this work here personally, seems deeper in this component stack than it needs to be, especially because this is the kind of work that would likely happen on the server.

jjm3x3 (approved):
LGTM

PR #480, opened 2026-03-22
Converted IContestantRoundData

internette:

I migrated IContestantRoundData both to the correct naming convention as well as to a type instead of interface. Refactoring the table code informed me that we need the type to be indexable, which is only in mappable types. Interfaces are not mappable. Learn more here: https://www.typescriptlang.org/docs/handbook/2/mapped-types.html

## Confirm
- [ ] This PR has unit tests scenarios. (N/A)
- [ ] This PR has been locally QA'd. (N/A)

### DIFF
--- a/app/models/ContestantRoundData.tsx
+++ b/app/models/ContestantRoundData.tsx
@@ -1,4 +1,4 @@
-export default interface IContestantRoundData {
+export type ContestantRoundData ={
     name: string
     roundScore: number
     totalScore: number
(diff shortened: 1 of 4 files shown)


app/models/ContestantRoundData.tsx
```diff
@@ -1,4 +1,4 @@
-export default interface IContestantRoundData {
+export type ContestantRoundData ={
```
jjm3x3: QQ: How come convert this from an `interface` to a `type`?

jjm3x3: Wait... I actually don't think that is true after doing my research 😅 _see [this comment](https://github.com/jjm3x3/AmazingRaceFantasy/pull/479#discussion_r2970733231), which is the conclusion_ 👀

PR #498, opened 2026-04-12
442 add league update endpoint

### LINKED ISSUE #442: Add league update endpoint
In order to complete the connection necessary to enable the post back on #439  we will need an api endpoint which will transitionally "update" the league config. If the status doesn't change and the session token has the same internal userId as the leagues createdBy userId. Of course if the status the the league is updated we will need to make sure the existing key is deleted and anew key added, but in a transaction so that we don't end up with 0 or 2 leagues, we can guarantee exactly one league. In typical REST fashion this endpoint should likely be a `PUT` and the endpoint should be `/api/configuration`

internette:
This creates a `PUT` endpoint for updating the league configuration. As discussed with Jacob, this new endpoint exists in the same file as `POST` under the league file. The main differences between the PUT and the POST, aside from method, is how we trigger the 403.

First we are getting the all league configuration keys based on data that's passed in. Then, we're doing the following checks:
- If there are zero keys found, throw a 404
- If there are more than one key found, throw a 500
- If there is exactly one key, fetch the data for the one key

When data fetching completes, we get the createdBy value. We use that to check the league creator's ID against the session ID, ensuring that the person submitting the change request is the owner and is allowed to make the change. If they are, then make the update.

## Confirm

### DIFF
--- a/__tests__/api/league/route.js
+++ b/__tests__/api/league/route.js
@@ -866,6 +866,23 @@ describe("PUT (unit tests)", () => {
 
     it("should return 200 when PUT conditions met", async () => {
         // Arrange
+        const dbWikiPageUrl = "https://en.wikipedia.org/wiki/Original_Page";
+        const dbWikiApiUrl = "https://en.wikipedia.org/w/api.php?action=parse&format=json&page=Original_Page";
+        const dbCastPhrase = "Original Cast";
+        const dbCompetitingEntityName = "person";
+        const dbContestantLeagueDataKeyPrefix = `${happyPathRequest.leagueKey}:*`;
+        
+        getLeagueConfigurationData.mockImplementation(() => {
+            return Promise.resolve({
+                createdBy: ourUserId,
+                wikiPageUrl: dbWikiPageUrl,
+                wikiApiUrl: dbWikiApiUrl,
+                castPhrase: dbCastPhrase,
+                competitingEntityName: dbCompetitingEntityName,
+                contestantLeagueDataKeyPrefix: dbContestantLeagueDataKeyPrefix
+            });
+        });
+
         const request = {
             cookies: {
                 get: jest.fn().mockImplementation(()=> {
@@ -873,11 +890,7 @@ describe("PUT (unit tests)", () => {
                 })
             },
             json: jest.fn().mockImplementation(async () => {
-                return {
-                    ...happyPathRequest,
-                    wikiPageUrl: `https://en.wikipedia.org/wiki/${happyPathRequest.wikiPageName}`,
-                    wikiApiUrl: `https://en.wikipedia.org/w/api.php?action=parse&format=json&page=${happyPathRequest.wikiPageName}`
-                };
+                return happyPathRequest;
             })
         };
 
@@ -890,9 +903,10 @@ describe("PUT (unit tests)", () => {
         expect(request.json).toHaveBeenCalledTimes(1);
         expect(writeLeagueConfigurationData).toHaveBeenCalledTimes(1);
         expect(writeLeagueConfigurationData).toHaveBeenCalledWith(
-            `league_configuration:${happyPathRequest.leagueStatus}:${happyPathRequest.leagueKey}`,
+            `league_configuration:${happyPathRequest.leagueStatus}:${dbContestantLeagueDataKeyPrefix}`,
             expect.objectContaining({
-                googleSheetUrl: happyPathRequest.googleSheetUrl
+                googleSheetUrl: happyPathRequest.googleSheetUrl,
+                leagueStatus: happyPathRequest.leagueStatus
             })
         );
     });
@@ -909,11 +923,7 @@ describe("PUT (unit tests)", () => {
                 })
             },
             json: jest.fn().mockImplementation(async () => {
-                return {
-                    ...happyPathRequest,
-                    wikiPageUrl: `https://en.wikipedia.org/wiki/${happyPathRequest.wikiPageName}`,
-                    wikiApiUrl: `https://en.wikipedia.org/w/api.php?action=parse&format=json&page=${happyPathRequest.wikiPageName}`
-                };
+                return happyPathRequest;
             })
         };
 
@@ -940,11 +950,7 @@ describe("PUT (unit tests)", () => {
                 })
             },
             json: jest.fn().mockImplementation(async () => {
-                return {
-                    ...happyPathRequest,
-                    wikiPageUrl: `https://en.wikipedia.org/wiki/${happyPathRequest.wikiPageName}`,
-                    wikiApiUrl: `https://en.wikipedia.org/w/api.php?action=parse&format=json&page=${happyPathRequest.wikiPageName}`
-                };
+                return happyPathRequest;
             })
         };
 
@@ -970,11 +976,7 @@ describe("PUT (unit tests)", () => {
                 })
             },
             json: jest.fn().mockImplementation(async () => {
-                return {
-                    ...happyPathRequest,
-                    wikiPageUrl: `https://en.wikipedia.org/wiki/${happyPathRequest.wikiPageName}`,
-                    wikiApiUrl: `https://en.wikipedia.org/w/api.php?action=parse&format=json&page=${happyPathRequest.wikiPageName}`
-                };
+                return happyPathRequest;
             })
         };
 
@@ -990,9 +992,7 @@ describe("PUT (unit tests)", () => {
         // Arrange
         const invalidRequestBody = {
             ...happyPathRequest,
-            leagueStatus: "maybe-active",
-            wikiPageUrl: `https://en.wikipedia.org/wiki/${happyPathRequest.wikiPageName}`,
-            wikiApiUrl: `https://en.wikipedia.org/w/api.php?action=parse&format=json&page=${happyPathRequest.wikiPageName}`
+            leagueStatus: "maybe-active"
         };
         const request = {
             cookies: {
@@ -1015,10 +1015,66 @@ describe("PUT (unit tests)", () => {
         expect(bodyString).toContain("leagueStatus");
     });
 
-    it("should use wikiPageUrl and wikiApiUrl from request body when updating", async () => {
+    it("should use wikiPageUrl and wikiApiUrl from database when updating", async () => {
         // Arrange
-        const customWikiPageUrl = "https://en.wikipedia.org/wiki/Custom_Page";
-        const customWikiApiUrl = "https://en.wikipedia.org/w/api.php?action=parse&format=json&page=Custom_Page";
+        const dbWikiPageUrl = "https://en.wikipedia.org/wiki/Database_Page";
+        const dbWikiApiUrl = "https://en.wikipedia.org/w/api.php?action=parse&format=json&page=Database_Page";
+        const dbContestantLeagueDataKeyPrefix = `${happyPathRequest.leagueKey}:*`;
+        
+        getLeagueConfigurationData.mockImplementation(() => {
+            return Promise.resolve({
+                createdBy: ourUserId,
+                wikiPageUrl: dbWikiPageUrl,
+                wikiApiUrl: dbWikiApiUrl,
+                castPhrase: "Cast",
+                competitingEntityName: "person",
+                contestantLeagueDataKeyPrefix: dbContestantLeagueDataKeyPrefix
+            });
+        });
+
+        const request = {
+            cookies: {
+                get: jest.fn().mockImplementation(()=> {
+                    return "testToken"
+                })
+            },
+            json: jest.fn().mockImplementation(async () => {
+                return happyPathRequest;
+            })
+        };
+
+        // Act
+        const response = await PUT(request);
+
+        // Assert
+        expect(response).not.toBeNull();
+        expect(response.status).toEqual(200);
+        expect(writeLeagueConfigurationData).toHaveBeenCalledWith(
+            expect.anything(),
+            expect.objectContaining({
+                wikiPageUrl: dbWikiPageUrl,
+                wikiApiUrl: dbWikiApiUrl
+            })
+        );
+    });
+
+    it("should use castPhrase and competitingEntityName from database when updating", async () => {
+        // Arrange
+        const dbCastPhrase = "Original Cast";
+        const dbCompetitingEntityName = "person";
+        const dbContestantLeagueDataKeyPrefix = `${happyPathRequest.leagueKey}:*`;
+        
+        getLeagueConfigurationData.mockImplementation(() => {
+            return Promise.resolve({
+                createdBy: ourUserId,
+                wikiPageUrl: "https://en.wikipedia.org/wiki/Page",
+                wikiApiUrl: "https://en.wikipedia.org/w/api.php",
+                castPhrase: dbCastPhrase,
+                competitingEntityName: dbCompetitingEntityName,
+                contestantLeagueDataKeyPrefix: dbContestantLeagueDataKeyPrefix
+            });
+        });
+
         const request = {
             cookies: {
                 get: jest.fn().mockImplementation(()=> {
@@ -1028,8 +1084,8 @@ describe("PUT (unit tests)", () => {
             json: jest.fn().mockImplementation(async () => {
                 return {
                     ...happyPathRequest,
-                    wikiPageUrl: customWikiPageUrl,
-                    wikiApiUrl: customWikiApiUrl
+                    castPhrase: "Different Cast",
+                    contestantType: "team"
                 };
             })
         };
@@ -1043,8 +1099,8 @@ describe("PUT (unit tests)", () => {
         expect(writeLeagueConfigurationData).toHaveBeenCalledWith(
             expect.anything(),
             expect.objectContaining({
-                wikiPageUrl: customWikiPageUrl,
-                wikiApiUrl: customWikiApiUrl
+                castPhrase: dbCastPhrase,
+                competitingEntityName: dbCompetitingEntityName
             })
         );
     });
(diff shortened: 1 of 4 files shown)


app/api/league/route.ts
```diff
@@ -83,3 +83,68 @@ export async function POST(request: NextRequest) {
     // return
     return NextResponse.json({"message": "posted"});
 }
+
+export async function PUT (request: NextRequest) {
+    // check auth
+    const body = await request.json();
+    const sessionCookie = request.cookies.get("session");
+    let userId: string | undefined;
+    if(sessionCookie){
```
jjm3x3: nit/fft: I would invert this condition so we can "fail fast" and we can actually remove the else I believe.

app/api/league/route.ts
```diff
@@ -83,3 +83,68 @@ export async function POST(request: NextRequest) {
...
+        const decryptedSessionCookie = await decrypt(sessionCookie?.value) as decryptionPayload;
+        userId = decryptedSessionCookie?.sub;
+        const invalidUserId = !userId;
+        const invalidUserSub = Object.keys(decryptedSessionCookie).length !== 3;
+        if (invalidUserId || invalidUserSub ) {
+            return NextResponse.json({"error": unauthenticatedErrorMessage}, {status: 401});
+        }
+        const leagueConfigurationKeyArray = await getAllKeys(`league_configuration:*:${body.leagueKey}`);
```
jjm3x3: We should not use the body before we have effectively parsed it.

app/api/league/route.ts
```diff
@@ -83,3 +83,68 @@ export async function POST(request: NextRequest) {
...
+        if(leagueConfigurationKeyArray.length === 0){
+            return NextResponse.json({"error": "no league configuration found for that league key"}, {status: 404});
+        } else if (leagueConfigurationKeyArray.length > 1){
+            return NextResponse.json({"error": "multiple league configurations found for that league key, please contact support"}, {status: 500});
```
jjm3x3: As we talked about 409, is probably better here.

jjm3x3 (commented):
These might be confusing, because I refreshed the page before submitting 😅

app/api/league/route.ts
```diff
@@ -83,3 +83,69 @@ export async function POST(request: NextRequest) {
...
+        }
+    }
+
+    // check permissions - only allow if user is league owner
+    const leagueConfigurationKeyArray = await getAllKeys(`league_configuration:*:${body.leagueKey}`);
+    if(leagueConfigurationKeyArray.length === 0){
+        return NextResponse.json({"error": "no league configuration found for that league key"}, {status: 404});
+    } else if (leagueConfigurationKeyArray.length > 1){
+        return NextResponse.json({"error": "multiple league configurations found for that league key, please contact support"}, {status: 409});
+    } else {
```
jjm3x3: nit: we don't need this `else` anymore, since we have the fail fasts in the other branches of the if.

app/api/league/route.ts
```diff
@@ -83,3 +83,69 @@ export async function POST(request: NextRequest) {
...
+        postGoogleSheetsLinkText: "which was populated using a google form.",
+        competitingEntityName: body.contestantType,
+        contestantLeagueDataKeyPrefix: `${body.leagueKey}:*`,
+        createdBy: userId
+    };
+
+    await writeLeagueConfigurationData(leagueConfigKey, leagueConfig);
+
+    // return
+    return NextResponse.json({"message": "posted"});
```
jjm3x3: nit/fft: should this instead say something like `"updated"`?

jjm3x3 (commented):
Few small nit's

app/api/league/route.ts
internette: Fixed

app/api/league/route.ts
internette: I like this. Updated!

app/api/league/route.ts
internette: Agreed. Updated

app/api/league/route.ts
internette: Agreed. Refactored to move the check above. Also the check now only includes editable values

app/api/league/route.ts
internette: Fixed.

__tests__/api/league/route.js
```diff
@@ -850,3 +854,408 @@ describe("POST (unit tests)", () => {
...
+        // Assert
+        expect(response).not.toBeNull();
+        expect(response.status).toEqual(200);
+        expect(request.json).toHaveBeenCalledTimes(1);
+        expect(writeLeagueConfigurationData).toHaveBeenCalledTimes(1);
+        expect(writeLeagueConfigurationData).toHaveBeenCalledWith(
+            `league_configuration:${happyPathRequest.leagueStatus}:${dbContestantLeagueDataKeyPrefix}`,
+            expect.objectContaining({
+                googleSheetUrl: happyPathRequest.googleSheetUrl,
+                leagueStatus: happyPathRequest.leagueStatus
```
jjm3x3: nit: How come we are only testing these two fields?

__tests__/api/league/route.js
```diff
@@ -850,3 +861,375 @@ describe("POST (unit tests)", () => {
...
+        expect(response.status).toEqual(200);
+        expect(request.json).toHaveBeenCalledTimes(1);
+        expect(writeLeagueConfigurationData).toHaveBeenCalledTimes(1);
+        expect(writeLeagueConfigurationData).toHaveBeenCalledWith(
+            `league_configuration:${happyPathRequest.leagueStatus}:${dbContestantLeagueDataKeyPrefix}`,
+            expect.objectContaining({
+                createdBy: happyPathRequest.createdBy,
+                leagueStatus: happyPathRequest.leagueStatus
+            })
+        );
```
jjm3x3: So we may want to have an in person conversation about this, but the short-ish version is this.

I don't think we actually want to call `writeLeagueConfigurationData` for the `PUT` here. And the reason is, because we would the possible write and delete to happen in the same transaction

__tests__/api/league/route.js
```diff
@@ -850,3 +861,375 @@ describe("POST (unit tests)", () => {
...
+        expect(response.status).toEqual(200);
+        expect(writeLeagueConfigurationData).toHaveBeenCalledWith(
+            expect.anything(),
+            expect.objectContaining({
+                wikiPageUrl: dbWikiPageUrl,
+                wikiApiUrl: dbWikiApiUrl,
+                castPhrase: dbCastPhrase,
+                competitingEntityName: dbCompetitingEntityName
+            })
+        );
```
jjm3x3: Similar to the comment above, not sure we actually want to assert this exact thing.

__tests__/api/league/route.js
```diff
@@ -850,3 +861,375 @@ describe("POST (unit tests)", () => {
...
+
+        // Act
+        const response = await PUT(request);
+
+        // Assert
+        expect(response).not.toBeNull();
+        expect(getAllKeys).toHaveBeenCalledWith(`league_configuration:*:${happyPathRequest.leagueKey}`);
+    });
+
+    it("should construct the write key using database contestantLeagueDataKeyPrefix", async () => {
```
jjm3x3: How come? I actually don't think this is true. 🤔

app/api/league/route.ts
```diff
@@ -83,3 +83,79 @@ export async function POST(request: NextRequest) {
...
+const LeagueConfigUpdate = z.object({
+    createdBy: validationPattern.createdBy.zod,
+    leagueStatus: validationPattern.leagueStatus.zod,
+    leagueKey: validationPattern.leagueKey.zod
+});
+
+export async function PUT (request: NextRequest) {
+    // check auth
+    const body = await request.json();
+    console.log(body);
```
jjm3x3: nit: probably don't need this.

```suggestion
```

app/api/league/route.ts
```diff
@@ -83,3 +83,79 @@ export async function POST(request: NextRequest) {
...
+    const userId = decryptedSessionCookie?.sub;
+    const invalidUserId = !userId;
+    const invalidUserSub = Object.keys(decryptedSessionCookie).length !== 3;
+    if (invalidUserId || invalidUserSub ) {
+        return NextResponse.json({"error": unauthenticatedErrorMessage}, {status: 401});
+    }
+
+    // validate/sanitize input
+    try {
+        LeagueConfigUpdate.parse(body);
```
jjm3x3: fft: We could add an interface here and some typing that would help catch even more mistakes going forward... something that looks a little like this:

```suggestion
    let updateRequest: UpdateRequest | null = null;
    try {
        LeagueConfigUpdate.parse(body);
        updateRequest = body as UpdateRequest;
```

app/api/league/route.ts
```diff
@@ -83,3 +83,79 @@ export async function POST(request: NextRequest) {
...
+    const preexistingLeagueConfigurationKey = leagueConfigurationKeyArray[0];
+    const leagueConfigurationData = await getLeagueConfigurationData(preexistingLeagueConfigurationKey);
+    const isUserDenied = userId !== leagueConfigurationData.createdBy;
+    if(isUserDenied){
+        return NextResponse.json({"error": "you are not authorized to perform that action"}, {status: 403});
+    }
+
+    if(leagueConfigurationData.leagueStatus !== body.leagueStatus){
+        await deleteLeagueConfigurationData(preexistingLeagueConfigurationKey);
+    }
```
jjm3x3: I do like this check, that way we can avoid extra DB calls if possible, however I think we should try to delete and add in a single transaction and possibly we should add then delete first. 

- [Here is the official Redis documentation on Transactions](https://redis.io/docs/latest/develop/using-commands/transactions/) _(though it is a bit dense)_
- [Here are the docs on the exact client library we are using for Upstash](https://upstash.com/docs/redis/sdks/ts/commands/transaction)

app/api/league/route.ts
```diff
@@ -83,3 +83,79 @@ export async function POST(request: NextRequest) {
...
+        leagueStatus: body.leagueStatus,
+        castPhrase: leagueConfigurationData.castPhrase,
+        preGoogleSheetsLinkText: "This season's contestant data has been sourced from",
+        postGoogleSheetsLinkText: "which was populated using a google form.",
+        competitingEntityName: leagueConfigurationData.competitingEntityName,
+        contestantLeagueDataKeyPrefix: leagueConfigurationData.contestantLeagueDataKeyPrefix,
+        createdBy: userId
+    };
+
+    await writeLeagueConfigurationData(leagueConfigKey, leagueConfig);
```
jjm3x3: Then, in case it goes w/o saying, we don't need this if we have built an "update" transaction.

jjm3x3 (dismissed):
Hate to do it, but I do really think we should do this in a transaction, plus it does say that in the issue description. Besides that a couple of much smaller suggestions and one minor bug, I believe.

That said this is very close.

internette:
> Hate to do it, but I do really think we should do this in a transaction, plus it does say that in the issue description. Besides that a couple of much smaller suggestions and one minor bug, I believe.
> 
> That said this is very close.

Do what in a transaction? I expressed my confusion in the discord and we will figure this out offline.

__tests__/api/league/route.js
internette: We aren't anymore. This is outdated.

__tests__/api/league/route.js
internette: We no longer do this. This has been updated and this comment is outdated.

__tests__/api/league/route.js
internette: This has been adjusted to assert appropriately.

__tests__/api/league/route.js
internette: This test has been corrected.

app/api/league/route.ts
internette: Done!

app/api/league/route.ts
internette: This has been changed and a PR was merged in to account for this

app/api/league/route.ts
internette: Correct, this has been updated

__tests__/api/league/route.js
```diff
@@ -850,3 +861,288 @@ describe("POST (unit tests)", () => {
...
+        expect(response.status).toEqual(200);
+        expect(updateLeagueConfigurationData).not.toHaveBeenCalledWith(
+            `${happyPathRequest.leagueStatus}:${dbContestantLeagueDataKeyPrefix}`,
+            expect.anything()
+        );
+        expect(getAllKeys.mock.calls[0][0]).toContain(`league_configuration:*:${happyPathRequest.leagueKey}`);
+        expect(getLeagueConfigurationData).toHaveBeenCalledWith(`league_configuration:${happyPathRequest.leagueStatus}:${happyPathRequest.leagueKey}`);
+        expect(updateLeagueConfigurationData).toHaveBeenCalledWith(
+            happyPathRequest.leagueKey,
+            expect.anything()
```
jjm3x3, line 1145: This should really not be from the request, right? This should be from the db.

app/api/league/route.ts
```diff
@@ -83,3 +90,74 @@ export async function POST(request: NextRequest) {
     // return
     return NextResponse.json({"message": "posted"});
 }
+
+const LeagueConfigUpdate = z.object({
+    createdBy: validationPattern.createdBy.zod,
+    leagueStatus: validationPattern.leagueStatus.zod,
+    leagueKey: validationPattern.leagueKey.zod
```
jjm3x3, line 97: Alright, we cannot merge this w/o at least including input validation on the google sheet.

jjm3x3: Along with validating this as input we also talked about this effectively being a required field which will always update the db accordingly. Given that, we should add a test (or at least an assertion on an existing test) to make sure that we are calling update with the google sheet url from the body.

app/api/league/route.ts
```diff
@@ -83,3 +90,74 @@ export async function POST(request: NextRequest) {
...
+    const body = await request.json();
+    const sessionCookie = request.cookies.get("session");
+    if(!sessionCookie){
+        return NextResponse.json({"error": unauthenticatedErrorMessage}, {status: 401});
+    }
+    
+    const decryptedSessionCookie = await decrypt(sessionCookie?.value) as decryptionPayload;
+    const userId = decryptedSessionCookie?.sub;
+    const invalidUserId = !userId;
+    const invalidUserSub = Object.keys(decryptedSessionCookie).length !== 3;
```
jjm3x3, line 111: This is on the wrong part of the credential. It is the `sessionCookie` that should be guaranteed to have 3 parts (separated by `.`s to be exact). Also this is also wrong in the `POST` method in the same controller. Probably worth fixing both. But we can fast follow with it.

app/api/league/route.ts
```diff
@@ -83,3 +90,74 @@ export async function POST(request: NextRequest) {
...
+
+    // check permissions - only allow if user is league owner
+    const leagueConfigurationKeyArray = await getAllKeys(`league_configuration:*:${body.leagueKey}`);
+    if(leagueConfigurationKeyArray.length === 0){
+        return NextResponse.json({"error": "no league configuration found for that league key"}, {status: 404});
+    } else if (leagueConfigurationKeyArray.length > 1){
+        return NextResponse.json({"error": "multiple league configurations found for that league key, please contact support"}, {status: 409});
+    }
+
+    const preexistingLeagueConfigurationKey = leagueConfigurationKeyArray[0];
```
jjm3x3, line 139: nit: is there not something like a `.first()` method that we can call, or maybe like a `.take(1)` which whould help describe intent that there is exactly one?

app/dataSources/dbFetch.tsx
```diff
@@ -95,6 +95,32 @@ export async function writeLeagueConfigurationData(leagueConfigurationKey: strin
...
+        throw new Error("Unable to writeLeagueConfigurationData. Provided param 'leagueConfiguration' is undefined but must have a value");
+    };
+
+    const redis = new Redis({
+        url: process.env.KV_REST_API_URL,
+        token: process.env.KV_REST_API_TOKEN
+    });
+
+    const leagueStatusAndKeyPrefix = leagueConfigurationKey.replace("league_configuration:", "");
+    const index = leagueStatusAndKeyPrefix.indexOf(":");
```
jjm3x3, line 112: nit: don't love the name index here 😅

app/dataSources/dbFetch.tsx
```diff
@@ -95,6 +95,32 @@ export async function writeLeagueConfigurationData(leagueConfigurationKey: strin
...
+    const leagueKeyWithoutStatus = leagueStatusAndKeyPrefix.slice(index + 1);
+    const leagueConfigString = JSON.stringify(leagueConfiguration);
+    const tx = redis.multi();
+    tx.del(leagueConfigurationKey);
+    const newLeagueConfigKey = `league_configuration:${leagueConfiguration.leagueStatus}:${leagueKeyWithoutStatus}`;
```
jjm3x3, line 117: nit: I would construct this before we do the delete, and outside the txn so that we keep that section as small as possible.

app/dataSources/dbFetch.tsx
```diff
@@ -113,6 +139,23 @@ export async function getLeagueConfigurationData(leagueConfigurationKey: string)
...
+        url: process.env.KV_REST_API_URL,
+        token: process.env.KV_REST_API_TOKEN
+    });
+    try {
+        await redis.del(leagueConfigurationKey);
+    } catch (error) {
+        throw new Error("There was an error deleting the league configuration data for the key provided");
+    }
+}
+
```
jjm3x3, line 158: Delete unused code:
```suggestion
```

PR #504, opened 2026-05-01
Convert to transaction

internette:
This updates the original update league config endpoint logic from individual functions to a transaction that combines delete and set. I removed the tests that test individual functions for write and delete in the PUT tests and didn't add more because those would be integration tests. We won't be testing Redis.

## Confirm
- [ ] This PR has been locally QA'd.

### DIFF
--- a/__tests__/api/league/route.js
+++ b/__tests__/api/league/route.js
@@ -873,8 +873,6 @@ describe("PUT (unit tests)", () => {
         getAllKeys.mockImplementation(() => {
             return Promise.resolve([`league_configuration:${happyPathRequest.leagueStatus}:${happyPathRequest.leagueKey}`]);
         });
-
-        
     });
 
     afterEach(()=> {
@@ -1102,20 +1100,9 @@ describe("PUT (unit tests)", () => {
         expect(getAllKeys).toHaveBeenCalledWith(`league_configuration:*:${happyPathRequest.leagueKey}`);
     });
 
-    it("should construct the write key using database contestantLeagueDataKeyPrefix", async () => {
+    it("should construct the update key using database contestantLeagueDataKeyPrefix", async () => {
         // Arrange
-        const dbContestantLeagueDataKeyPrefix = `${happyPathRequest.leagueKey}:*`;
-        getLeagueConfigurationData.mockImplementation(() => {
-            return Promise.resolve({
-                createdBy: ourUserId,
-                leagueStatus: happyPathRequest.leagueStatus,
-                wikiPageUrl: "https://en.wikipedia.org/wiki/Page",
-                wikiApiUrl: "https://en.wikipedia.org/w/api.php",
-                castPhrase: "Cast",
-                competitingEntityName: "person",
-                contestantLeagueDataKeyPrefix: dbContestantLeagueDataKeyPrefix
-            });
-        });
+        const dbContestantLeagueDataKeyPrefix = "tvshowTestKey";
 
         const request = {
             cookies: {
@@ -1124,9 +1111,25 @@ describe("PUT (unit tests)", () => {
                 })
             },
             json: jest.fn().mockImplementation(async () => {
-                return happyPathRequest;
+                return {
+                    createdBy: ourUserId,
+                    leagueStatus: "active",
+                    leagueKey: happyPathRequest.leagueKey
+                }
             })
         };
+
+        getLeagueConfigurationData.mockImplementation(() => {
+            return Promise.resolve({
+                createdBy: ourUserId,
+                leagueStatus: happyPathRequest.leagueStatus,
+                wikiPageUrl: "https://en.wikipedia.org/wiki/Page",
+                wikiApiUrl: "https://en.wikipedia.org/w/api.php",
+                castPhrase: "Cast",
+                competitingEntityName: "person",
+                contestantLeagueDataKeyPrefix: `${dbContestantLeagueDataKeyPrefix}:*`
+            });
+        });
 
         // Act
         const response = await PUT(request);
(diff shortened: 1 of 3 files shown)


app/dataSources/dbFetch.tsx
```diff
@@ -95,6 +95,28 @@ export async function writeLeagueConfigurationData(leagueConfigurationKey: strin
...
+    const redis = new Redis({
+        url: process.env.KV_REST_API_URL,
+        token: process.env.KV_REST_API_TOKEN
+    });
+
+    const leagueConfigString = JSON.stringify(leagueConfiguration);
+    const tx = redis.multi();
+    if(tx.exists(leagueConfigurationKey)){
+        tx.del(leagueConfigurationKey);
+    }
```
jjm3x3: Do we need to call `tx.exists` to check for the existence of the key?

app/dataSources/dbFetch.tsx
```diff
@@ -95,6 +95,28 @@ export async function writeLeagueConfigurationData(leagueConfigurationKey: strin
     await redis.json.set(leagueConfigurationKey, "$", leagueConfigString)
 }
 
+export async function updateLeagueConfigurationData(leagueConfigurationKey: string, leagueConfiguration: ILeagueConfigurationData): Promise<void> {
+    if (leagueConfigurationKey === undefined) {
+        throw new Error("Unable to writeLeagueConfigurationData. Provided param 'leagueConfigurationKey' is undefined but must have a value\"");
+    };
+    if (leagueConfiguration === undefined) {
+        throw new Error("Unable to writeLeagueConfigurationData. Provided param 'leagueConfiguration' is undefined but must have a value\"");
```
jjm3x3: nit: please remove extra quote.

```suggestion
        throw new Error("Unable to writeLeagueConfigurationData. Provided param 'leagueConfiguration' is undefined but must have a value");
```

jjm3x3 (changes requested):
We talked about it offline, but we need to make sure that we have enough info both old and new key to perform the update.

__tests__/api/league/route.js
```diff
@@ -1140,96 +1134,9 @@ describe("PUT (unit tests)", () => {
         // Assert
         expect(response).not.toBeNull();
         expect(response.status).toEqual(200);
-        expect(writeLeagueConfigurationData).toHaveBeenCalledWith(
-            `league_configuration:${happyPathRequest.leagueStatus}:${dbContestantLeagueDataKeyPrefix}`,
```
jjm3x3, line 1144: nit: I feel like we should make this a value that highlights the purpose of this test a bit better.

jjm3x3: I also think we should do another test like  this, but for the `:<status>:` portion of the key.

jjm3x3 (approved):
A few small comments.

__tests__/api/league/route.js
jjm3x3: Alright, we talked about this and we actually did confirm that it should come from the request (at least for the last part of the key). 

But because that should be the same as the end of they key we get back from `getAllKeys` and we have the test called `should call getAllKeys with correct league key pattern`

So for this test, we can just update the assertion to highlight that this is the existing/old key that came from our `getAllKeys` call. And then we don't need another test to assert on the `:<status>:` segment.

app/dataSources/dbFetch.tsx
```diff
@@ -95,6 +95,28 @@ export async function writeLeagueConfigurationData(leagueConfigurationKey: strin
...
+    const redis = new Redis({
+        url: process.env.KV_REST_API_URL,
+        token: process.env.KV_REST_API_TOKEN
+    });
+    
+
+    const leagueConfigString = JSON.stringify(leagueConfiguration);
+    const tx = redis.multi();
+    tx.del(leagueConfigurationKey);
+    const newLeagueConfigKey = `league_configuration:${leagueConfiguration.leagueStatus}:${leagueConfiguration.contestantLeagueDataKeyPrefix}`;
```
jjm3x3: The last argument here should come from the request as we discussed in person. Unfortunately the only access we have to that data in this method is by extracting everything after the last `':'` in the passed in `leagueConfigurationKey`

app/dataSources/dbFetch.tsx
internette: No we don't. Turns out that redis.del is idempotent

app/dataSources/dbFetch.tsx
jjm3x3: fft: Also I think rightfully this code does start to feel like it's doing a bit more than a simple integration layer that we can mock out and so it might be worth considering a more robust middle layer where this logic can be tested a bit better. 

Of course doing so also begs a different question of should the idea of it being a transaction be moved up a layer and how does that abstraction work so probably not a change for right now, but something to think a little more about.

jjm3x3 (changes requested):
_Sorry_... as we discussed in person, this actually isn't quite right, there is at least one more change we need to make in the `dbFetch` function.

__tests__/api/league/route.js
jjm3x3: We again, talked about his offline and you didn't address any of this feedback 😅

jjm3x3 (approved):
Alright, except the one-ish piece of feedback on testing, the rest of this looks good. Feel free to merge once you feel like you have got that on lock.

__tests__/api/league/route.js
internette: Ok, I've updated this test to ensure getAllKeys, getLeagueConfigurationData, and updateLeagueConfigurationData all use the same body.leagueKey. I can't compare getAllKeys to updateLeagueConfigurationData directly because getAllKeys has a * in place of the league status, but ensuring the same league key is used should be sufficient

app/dataSources/dbFetch.tsx
internette: So, I went ahead and implemented the extraction:
```js
const leagueStatusAndKeyPrefix = leagueConfigurationKey.replace("league_configuration:", "");
    const index = leagueStatusAndKeyPrefix.indexOf(":");
    const leagueKeyWithoutStatus = leagueStatusAndKeyPrefix.slice(index + 1);
    const leagueConfigString = JSON.stringify(leagueConfiguration);
    const tx = redis.multi();
    tx.del(leagueConfigurationKey);
    const newLeagueConfigKey = `league_configuration:${leagueConfiguration.leagueStatus}:${leagueKeyWithoutStatus}`;
    tx.json.set(newLeagueConfigKey, "$", leagueConfigString);
    await tx.exec();
```

